PR #192, opened 2012-06-28
Add the streaming indexer to the bindings

carlosmn:
There are a few things that I'm not so sure about, style-wise.

Firstly, the indexer's stats are actually `unsigned int` in C, but if I try to put that in the struct, I get a warning that it's not CLS complaint, so it it may not work. @nulltoken, do you have any insights into this?

Secondly, I transform the stats into the C# variant after each add, which could be too much work, if the stats are looked up less often. I thought about doing it on demand, but that would mean locking and I worry that it would mean even more work.

Thirdly, is there some extablished convention whereby I can make clear that you shouldn't call `Indexer.Hash()` before the indexer is done?

### DIFF
--- a/LibGit2Sharp/Advanced/Indexer.cs
+++ b/LibGit2Sharp/Advanced/Indexer.cs
@@ -6,68 +6,76 @@ using LibGit2Sharp.Core.Handles;
 
 namespace LibGit2Sharp.Advanced
 {
+    /// <summary>
+    /// The Indexer is our implementation of the git-index-pack command. It is used to process the packfile
+    /// which comes from the remote side on a fetch in order to create the corresponding .idx file.
+    /// </summary>
     public class Indexer : IDisposable
     {
 
         readonly IndexerSafeHandle handle;
         readonly TransferProgressHandler callback;
 
-        GitTransferProgress progress;
-        byte[] buffer;
-
-        /// <summary>
-        /// The indexing progress
-        /// </summary>
-        /// <value>The progres information for the current operation.</value>
-        public TransferProgress Progress
-        {
-            get
-            {
-                return new TransferProgress(progress);
-            }
-        }
-
-        public Indexer(string prefix, uint mode, ObjectDatabase odb = null, TransferProgressHandler onProgress = null)
+        Indexer(string prefix, uint mode, ObjectDatabase odb = null, TransferProgressHandler onProgress = null)
         {
             /* The runtime won't let us pass null as a SafeHandle, wo create a "dummy" one to represent NULL */
             ObjectDatabaseSafeHandle odbHandle = odb != null ? odb.Handle : new ObjectDatabaseSafeHandle();
             callback = onProgress;
             handle = Proxy.git_indexer_new(prefix, odbHandle, mode, GitDownloadTransferProgressHandler);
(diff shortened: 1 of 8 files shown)


carlosmn:
I've brought this back from the dead, the style issues have been addressed and we only actually deal with `Stream`s anymore.

I'm not sure how much explanation there should be for the class; I figured mentioning it's the equivalent to index-pack should be enough for anybody who would want to use it.

LibGit2Sharp/Advanced/Indexer.cs
```diff
@@ -0,0 +1,109 @@
...
+            int read;
+            var indexProgress = default(GitTransferProgress);
+
+            using (var idx = new Indexer(prefix, mode, odb, onProgress))
+            {
+                var handle = idx.handle;
+
+                do
+                {
+                    read = stream.Read(buffer, 0, buffer.Length);
```
ethomson: `Stream.Read` can return negative here on failure - we should break if `read <= 0`.

LibGit2Sharp/Advanced/Indexer.cs
```diff
@@ -0,0 +1,109 @@
...
+        /// <param name="progress">The amount of objects processed etc will be written to this structure on exit</param>
+        /// <param name="path">Path to the file to index</param>
+        /// <param name="prefix">Path in which to store the pack and index files</param>
+        /// <param name="mode">Filemode to use for creating the pack and index files</param>
+        /// <param name="odb">Optional object db to use if the pack contains thin deltas</param>
+        /// <param name="onProgress">Function to call to report progress. It returns a boolean indicating whether
+        /// to continue working on the stream</param>
+
+        /// </summary>
+        public static ObjectId Index(out TransferProgress progress, string path, string prefix, uint mode, ObjectDatabase odb = null, TransferProgressHandler onProgress = null)
```
ethomson: We should have proper overloads here instead of optional params.

LibGit2Sharp/Advanced/Indexer.cs
```diff
@@ -0,0 +1,109 @@
...
+        /// in a background thread.
+        /// </summary>
+        /// <param name="progress">The amount of objects processed etc will be written to this structure on exit</param>
+        /// <param name="stream">Stream to run the indexing process on</param>
+        /// <param name="prefix">Path in which to store the pack and index files</param>
+        /// <param name="mode">Filemode to use for creating the pack and index files</param>
+        /// <param name="odb">Optional object db to use if the pack contains thin deltas</param>
+        /// <param name="onProgress">Function to call to report progress. It returns a boolean indicating whether
+        /// to continue working on the stream</param>
+        public static ObjectId Index(out TransferProgress progress, Stream stream, string prefix, uint mode, ObjectDatabase odb = null, TransferProgressHandler onProgress = null)
```
ethomson: We should have proper overloads here instead of optional params.

ethomson:
Let's get this out of `LibGit2Sharp.Advanced`.  Instead, let's just put it into `LibGit2Sharp`...  This seems to be in keeping with the trend of making `LibGit2Sharp` the "core" area and move the "simpler" stuff into `Commands`...?

PR #345, opened 2013-02-20
Tests for case sensitive conflicts in the index

ethomson:
So after looking at https://github.com/libgit2/libgit2sharp/pull/344 , I wanted to add a test case that had conflicts in the index that differed only in case.

core git will allow a merge of two branches that contain items differ in case, regardless of your platform or your `core.ignorecase` settings and we should, I would think, be able to handle those conflicts.

My thought was that these tests would simply pass against the current code base, but to my surprise, they do not...!  I have not yet been able to debug this, but it appears that `ignorecase` is a deeper setting than I had imagined.

I did want to push up this test, though, to illustrate a problem that can arise and start some discussion.

/cc @nulltoken
/cc @dahlbyk


carlosmn:
The C version of the test has been merged, making adding conflicts align with adding hand-crafted index entries in that we take whatever string was passed in, potentially changing the existing case in case-insensitive indices.

Do we still want to do something here?

carlosmn:
Since this changes so many files marked as binary and we did merge the C test, I'm going to close this as resolved and we'll let the library say what's right in this.

Issue #387, opened 2013-04-08
Encoding issue with commit name


Haacked:
Not sure if this is a LibGit2 or LibGit2Sharp bug but thought I'd start here. :)

Try cloning this repository here: https://github.com/CoolDevs/Mets-palopeli

```
git show d527f20
```

Shows the author as: Lassi Hämäläinen

But if you load the commit in LibGit2Sharp, you get: Lassi H�m�l�inen

Phil

Haacked:
> @Haacked Do you have two cloned repositories? I see that some commits have good encoding and some have bad. It is weird as they are commited in short time span with good and bad encoding.

Sorry I missed this all these years. :stuck_out_tongue: 

This is not my repository, so I don't know much about the context for why there are good and bad commits.

ethomson:
We talked about this in-person at Git Merge 2015 (wow, this issue is old) and decided that we would not try to do any fallbacks, as they would almost always be incorrect on the host machine.

PR #427, opened 2013-05-09
Implement encoding fallback for commit data

martinwoodward:
Work-around to resolve issue #387 where a repo has commit data that is not correctly encoded as UTF8. Not sure that this should be merged yet but throwing it out there for comment.


ethomson:
This is quite stale at this point.  Closing it.

PR #177, opened 2013-07-11
Add support for "git checkout <sha> -- file"

jamill:
Add support for the Git command:

`git checkout <sha> -- file`

For example, to checkout `foo.txt` from the changeset `HEAD`, it would be:

`git checkout HEAD -- foo.txt`

It would be nice if the API also exposed a parameter to specify the output path (i.e. the path the file is checked out to, which by default is the path the file is on the revision).

This requires handling of line endings (CRLF, CR and LF conversions).


jlindner:
Any chance CheckoutOptions will support sparseCheckout anytime soon?

PR #482, opened 2013-08-07
Repository.Submodules.Add created

olmobrutall:
This is my first pull request. I've created an Add method to Submodules. I haven't been able to run xUnit on VS so there's no test. 

Feel free to change it.


JasonGoemaat:
I wouldn't mind taking this over as I need this functionality for a project I'm working on converting from another source control provider to git.  I just have one issue that I'm not sure about.  

I can't find in the libgit2 api how you can add a submodule specifying a name that is different than the path:

	GIT_EXTERN(int) git_submodule_add_setup(
		git_submodule **out,
		git_repository *repo,
		const char *url,
		const char *path,
		int use_gitlink);

Does it make sense then for SubmoduleCollection.Add() to only take the path and not the name then?  The submodule ends up getting named whatever the relative path is.

carlosmn:
This is not functionality that belongs in `SubmoduleCollection` which should be about managing them, rather than emulating git's higher-level commands.

As this rather old, I'm going to close it, but if you still want to, something like this would belong in `Commands`.

PR #672, opened 2014-04-10
Extending repo.Network.Pull() #671

sitereactor:
Trying to help out with an up-for-grabs PR for issue #671 :smile:

I've added both Branch and Remote as an overload to Pull, but the Branch isn't really used for anything except checking if its Tracking. So not entirely sure if it makes sense to have or if it should be limited to Remote ?

Anyway, happy to change based on feedback.

Thanks!

### DIFF
--- a/LibGit2Sharp/Network.cs
+++ b/LibGit2Sharp/Network.cs
@@ -347,6 +347,34 @@ namespace LibGit2Sharp
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


carlosmn:
Thanks, but I don't think this is a direction we want to go. `git pull` doesn't take in branch names but refspecs and we should not be in the business of making up semantics for this. Something like this for `Commands.Pull()` that takes in a refspec (or multiple) is what we would want for exposing extra functionality in our pull.

PR #684, opened 2014-04-22
Submodule add

pms1969:
just added to the work of olmobrutall.  tests included.


carlosmn:
This is not functionality that belongs in `SubmoduleCollection` which should be about managing them, rather than emulating git's higher-level commands.

As this rather old, I'm going to close it, but if you still want to, something like this would belong in `Commands`.

PR #778, opened 2014-07-08
Nuget improvements

Zoltu:
* Adds a new project to the solution for building a native NuGet package.
 * The package ends up in `bin\<configuration>`.
 * When the package is installed, a build target file is imported into the target project that will cause the binaries to be copied to `$(TargetPath)` after build (if necessary).
 * Versioned as the last version of the LibGit2 that was pulled into this repository.
 * Do not be fooled, this is not actually a .csproj. In order to get Visual Studio to load it so that it shows up in Solution Explorer you have to tag it as one of a known list of project file types (http://msdn.microsoft.com/en-us/library/vstudio/ms171468(v=vs.110).aspx).  If you try to view the properties on it, Visual Studio will add a bunch of stuff to the underlying project file.  I don't *think* anything will break, but if it adds back the imports I believe it will break.  Be aware, best to just not open the properties on the project and instead open it in a text editor.

* Changes NuGet package to be generated on successful build of the class library.
 * The package ends up in `bin\<configuration>`.
 * Depends on the Native NuGet package (see above).
 * Only installs the one library and the one dependency, nothing fancy.

* Adds dependency on `NuGet for MSBuild`.
 * Both projects have a `.targets` file added to them that utilizes `NuGet for MSBuild`.
 * Take care in updating `NuGet for MSBuild`, it will add some post-build tasks to your project when you install/update it.  I have removed these but they will come back if you update.  They may have undesired side effects.
 * I have committed the dependency to the repo, I would encourage you (project owners) to `.gitignore` `packages/` though since Visual Studio will restore them on first build automatically (since VS2012 Update ? I believe).


Zoltu:
Recreated PR off vNext.  It was too much trouble to rebase/merge this one, so I just created a new branch.  New PR can be found here: #896.

PR #854, opened 2014-10-29
Register custom filters

ammeep:
I want to expose an API for registering custom clean and smudge filters via libgit2.

My knowledge of C is rusty, and this is the first time I have looked at libgit2sharp in any great detail. So, I'm going to open this PR early, in the hopes that if anyone has time to spot anything terrible I do along the way they will be able to point me in the right direction :pray: 

###  Todo

- [x] Register and deregsiter a filter
- [x] Define apply callback
- [x] Define check callback
- [x] Define shutdown/init callbacks
- [x] Tests


LibGit2Sharp/GlobalSettings.cs
```diff
@@ -108,5 +109,30 @@ public static LogConfiguration LogConfiguration
...
             }
         }
+
+        /// <summary>
+        /// Register a filter globally
+        /// </summary>
+        public static void RegisterFilter(Filter filter)
+        {
+            var nativeFilter = new GitFilterSafeHandle(filter.ManagedFilter);
+            Proxy.git_filter_register(filter.Name, nativeFilter, 0);
```
vmaha: Should this be `GIT_FILTER_DRIVER_PRIORITY`?

    /**
     * This is priority to use with a custom filter to imitate a core Git
     * filter driver, so that it will be run last on checkout and first on
     * checkin.  You do not have to use this, but it helps compatibility.
     */
    #define GIT_FILTER_DRIVER_PRIORITY 200

[Here's the definition](https://github.com/libgit2/libgit2/blob/091165c53b2bcd5d41fb71d43ed5a23a3d96bf5d/include/git2/sys/filter.h)

LibGit2Sharp/Core/Proxy.cs
```diff
@@ -745,6 +745,50 @@ public static GitDiffDelta git_diff_get_delta(DiffSafeHandle diff, int idx)
...
+        #region git_filter_
+
+        public static void git_filter_register(string name, GitFilterSafeHandle filter, int priority)
+        {
+            using (ThreadAffinity())
+            {
+                int res = NativeMethods.git_filter_register(name, filter, priority);
+                if (res == (int)GitErrorCode.Exists)
+                {
+                    var message = string.Format("A filter with the name '{0}' is already registered.", name);
```
vmaha: Nit: I'm new here so I'm not sure how much consistency matters, but I noticed you've added a `.`  to the end of this string, but the error message in `git_transport_register`, which is similar to this method, doesn't have a `.`.

LibGit2Sharp.Tests/TestHelpers/SubstitutionCipherFilter.cs
```diff
@@ -0,0 +1,58 @@
...
+        public int SmudgeCalledCount = 0;
+
+        public SubstitutionCipherFilter(string name, string attributes)
+            : base(name, attributes)
+        {
+        }
+
+        protected override int Check(IEnumerable<string> attributes, FilterSource filterSource)
+        {
+            CheckCalledCount++;
```
nulltoken: Can't we find a way to make this generic enough to push it down into the base class?

LibGit2Sharp/Core/GitBufWriter.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// </summary>
+        /// <param name="bytes">The bytes to write</param>
+        public override void Write(byte[] bytes)
+        {
+            IntPtr reverseBytesPointer = Marshal.AllocHGlobal(bytes.Length);
+            Marshal.Copy(bytes, 0, reverseBytesPointer, bytes.Length);
+
+            var size = (UIntPtr)bytes.LongLength;
+            var allocatedSize = (UIntPtr)bytes.LongLength + 1;
+            NativeMethods.git_buf_set(gitBuf, reverseBytesPointer, size);
```
nulltoken: Unless I'm wrong, every call to `Write()` will replace the content of the buffer instead of appending to it.

LibGit2Sharp/Filter.cs
```diff
@@ -0,0 +1,296 @@
...
+        /// <summary>
+        /// Shutdown callback on filter
+        /// 
+        /// Specified as `filter.shutdown`, this is an optional callback invoked
+        /// when the filter is unregistered or when libgit2 is shutting down.  It
+        /// will be called once at most and should release resources as needed.
+        /// This may be called even if the `initialize` callback was not made.
+        /// Typically this function will free the `git_filter` object itself.
+        /// </summary>
+        protected virtual void ShutDown()
```
nulltoken: How about trying to expose a more .Net friendly approach?

We could allow the Filter derived types to be `IDisposable`.
`GlobalSettings.Deregister()` would call `Dispose()` is the type is indeed `IDisposable`.

dahlbyk: :ghost: says :+1:

LibGit2Sharp.Tests/TestHelpers/SubstitutionCipherFilter.cs
dahlbyk: Not sure what you mean by "generic enough"?

LibGit2Sharp/Core/Proxy.cs
dahlbyk: > I'm new here so I'm not sure how much consistency matters (especially for dev-facing strings like exception messages)...

This is not the first feedback given about punctuation, nor will it be the last. You're in good company.

LibGit2Sharp/Filter.cs
```diff
@@ -0,0 +1,296 @@
...
+        {
+            return 0;
+        }
+
+        /// <summary>
+        /// Decides if a given source needs to be filtered.
+        /// </summary>
+        /// <param name="filterForAttributes">The filterForAttributes that this filter was created for.</param>
+        /// <param name="filterSource">The source of the filter</param>
+        /// <returns>0 if successful and -30 to skip and pass through</returns>
```
dahlbyk: Rather than `0` or `-30`, would it make sense to model this at the lg2s level as returning `true` (needs to be filtered) or `false` (does not)?

LibGit2Sharp/Core/GitBufWriter.cs
dahlbyk: I don't think you're wrong. More generally, reflecting `BinaryWriter` shows its methods are just helpers for writing to an internal `OutStream`:

```csharp
    public virtual void Write(byte[] buffer)
    {
      if (buffer == null)
        throw new ArgumentNullException("buffer");
      this.OutStream.Write(buffer, 0, buffer.Length);
    }
```

Attempting to use any `Write()` overload other than `Write(byte[])` will result in unexpected behavior.

It seems to me that we just need a solid `GitBufStream` implementation against which any writer (`BinaryWriter`, `TextWriter`, etc) — or reader, for that matter — would work.

dahlbyk:
Following on from [this comment](https://github.com/libgit2/libgit2sharp/pull/854#discussion_r23895973), it seems to me that both `Clean` and `Smudge` should have a pair of `Stream` parameters rather than us-specific binary reader/writer. Thoughts?

nulltoken:
>Following on from [this comment](https://github.com/libgit2/libgit2sharp/pull/854#discussion_r23895973), it seems to me that both `Clean` and `Smudge` should have a pair of `Stream` parameters rather than us-specific binary reader/writer. Thoughts?

I'd also definitively prefer this approach as well.

LibGit2Sharp.Tests/TestHelpers/SubstitutionCipherFilter.cs
nulltoken: >Not sure what you mean by "generic enough"?

Sorry, I was unclear. I was wondering if there was a real need to allow the end user to provide the system with his own `Check()` implementation. Is this sounds like a achievable goal, is this implementation enough?

LibGit2Sharp/Core/GitBufWriter.cs
ammeep: I think I was foolish to make GitBufWriter and readers inherit from anything that was streaming related. 

>It seems to me that we just need a solid GitBufStream implementation against which any writer 

The underlying libgit2 api does not get support streaming, so if we were to do something like this, half of the implementation would have to throw. Currently there is only one way to write back to the underlying buffer in libgit2, and it unfortunatly requires you to allocate the entire block.

You can follow the libgit2 discussion here https://github.com/libgit2/libgit2/issues/2757

I think until that lands, we are best to just remove any mention of streaming. Thoughts?

ammeep:
See the comment above about the underlying api not being compatible with streamed writes back to the filter registry. It's unfortunate, but for now I think we should remove any reference to 'Streams' until we can interop though libgit2 to do this.

LibGit2Sharp/Filter.cs
ammeep: These protected methods are actually wired up as callbacks for libgit2 so returning an integer is required. These methods are not designed to be called from user code at all.

We could however wrap these methods with one that translates the 0 or -30 result with a true or false, but that would fail to account for libgit2s support of filters returning custom error codes.

LibGit2Sharp/Filter.cs
ammeep: I'm pretty sure this will work just fine. I will change it to not register a shutdown callback with libgit2. This might also mean can we get rid of the clean call back too.

carlosmn:
I'd argue there's value in providing a spill buffer for the user even if libgit2 does not support this right now (I want to bring it back, but can't make promises about time). Adding the filters is enough work for a PR, but it might be something to do once this has settled (as it's bound to be faster than doing it in C).

ammeep:
That sounds like a good idea @carlosmn :sparkles:

LibGit2Sharp/GlobalSettings.cs
ammeep: Good spot @vmaha!

LibGit2Sharp.Tests/TestHelpers/SubstitutionCipherFilter.cs
ammeep: I've opted for a default implementation like we see here, with the ability to still override with your own behaviour if you need to :+1:

ammeep:
@nulltoken @dahlbyk Thanks for being patient while I went dark after your last round of feedback :sparkles: I got swamped the last couple of weeks.

LibGit2Sharp/GlobalSettings.cs
```diff
@@ -108,5 +109,40 @@ public static LogConfiguration LogConfiguration
...
+        /// Register a filter globally
+        /// </summary>
+        public static void RegisterFilter(Filter filter)
+        {
+            var nativeFilter = new GitFilterSafeHandle(filter.ManagedFilter);
+
+            // Register the filter with a priority of 200 allowing the custom filter 
+            // to imitate a core Git filter driver. 
+            // It will be run last on checkout and first on checkin.
+            Proxy.git_filter_register(filter.Name, nativeFilter, 200);
```
dahlbyk: Could this be introduced as a (documented) parameter with 200 as a (documented) default value? This is already a low-level feature - is there a reason not to expose this?

LibGit2Sharp.Tests/TestHelpers/SubstitutionCipherFilter.cs
```diff
@@ -0,0 +1,58 @@
...
+            var fileInfo = new FileInfo(filterSource.Path);
+            var matches = attributes.Any(currentExtension => fileInfo.Extension == currentExtension);
+            return matches ? 0 : -30;
```
dahlbyk: Delegate to `base.Check()` for default implementation?

LibGit2Sharp.Tests/TestHelpers/SubstitutionCipherFilter.cs
```diff
@@ -0,0 +1,58 @@
...
+            return RotatByThirteenPlaces(input, output);
+        }
+
+        protected override int Smudge(string path, GitBufReader input, GitBufWriter output)
+        {
+            SmudgeCalledCount++;
+            return RotatByThirteenPlaces(input, output);
+        }
+
+        public static int RotatByThirteenPlaces(GitBufReader input, GitBufWriter output)
```
dahlbyk: "Rotate" :smiling_imp:

dahlbyk:
> See the comment above about the underlying api not being compatible with streamed writes back to the filter registry. It's unfortunate, but for now I think we should remove any reference to Streams until we can interop though libgit2 to do this.

New territory for me, so I took a crack at `Stream`-ifying while keeping the current write-all-at-once implementation: https://github.com/libgit2/libgit2sharp/compare/register-custom-filters...dahlbyk/register-custom-filters. I could open a PR into this PR for discussion or...?

LibGit2Sharp/GlobalSettings.cs
ammeep: I think this is a great shout. It describe the order in which your filter is run, which has potential implications for a custom implementation of a filter. I'll expose it :+1:

LibGit2Sharp/GlobalSettings.cs
```diff
@@ -108,5 +109,46 @@ public static LogConfiguration LogConfiguration
...
+        /// Register a filter globally with a default priority of 200 allowing the custom filter 
+        /// to imitate a core Git filter driver. It will be run last on checkout and first on checkin.
+        /// </summary>
+        public static void RegisterFilter(Filter filter)
+        {
+            RegisterFilter(filter, 200);
+        }
+        /// <summary>
+        /// Register a filter globally with given priority for execution.
+        /// A filter with the priority of 200 It will be run last on checkout and first on checkin.
```
dahlbyk: "...200 will be..."

ammeep:
@dahlbyk this looks great - it gives us the spill buffer @carlosmn mentioned. Since you have done some hard work it would make sense to open this up for discussion :sparkles:

LibGit2Sharp/Filter.cs
```diff
@@ -0,0 +1,253 @@
...
+            : this(name, string.Join(",", attributes)) 
+        { }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Filter"/> class.
+        /// And allocates the filter natively. 
+        /// <param name="name">The unique name with which this filtered is registered with</param>
+        /// <param name="attributes">Either a single attribute, or a comma separated list of filterForAttributes for which this filter applies to</param>
+        /// </summary>
+        protected Filter(string name, string attributes)
```
nulltoken: How about not make this ctor accessible to derived types and favor the first one?

LibGit2Sharp/Filter.cs
```diff
@@ -0,0 +1,253 @@
...
+
+        /// <summary>
+        /// Initialize callback on filter
+        /// 
+        /// Specified as `filter.initialize`, this is an optional callback invoked
+        /// before a filter is first used.  It will be called once at most.
+        /// 
+        /// If non-NULL, the filter's `initialize` callback will be invoked right
+        /// before the first use of the filter, so you can defer expensive
+        /// initialization operations (in case libgit2 is being used in a way that doesn't need the filter).
```
nulltoken: libgit2 -> the library ?

LibGit2Sharp/FilterSource.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        /// <summary>
+        /// The relative path to the file
+        /// </summary>
+        public virtual string Path { get; private set; }
+
+        /// <summary>
+        /// A pointer to the repository
+        /// </summary>
+        public virtual IntPtr RepositoryHandle { get; private set; }
```
nulltoken: Do we need this exposed?

LibGit2Sharp/GlobalSettings.cs
```diff
@@ -108,5 +109,46 @@ public static LogConfiguration LogConfiguration
...
+            if (disposableFilter != null)
+            {
+                disposableFilter.Dispose();
+            }
+        }
+
+        /// <summary>
+        /// Remove the filter from the registry, and frees the native heap allocation.
+        /// </summary>
+        public static void DeregisterFilter(string name)
```
nulltoken: Is there a compelling reason to keep this overload? I'd rather enforce the usage of the one above, if possible.

ammeep: Simply that in an app with a long life time, you simply need to keep around the name of the filter, rather than an instance of it so you can deregister on shut down. 

It makes sense that if your filter is allocating things which need to be disposed of at the end of the filter life time that you would keep a reference to the filter around. But in the case where you do not this convenience (deregister) by name only seems useful.

LibGit2Sharp/Filter.cs
ammeep: :+1:

LibGit2Sharp/Filter.cs
```diff
@@ -0,0 +1,253 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Filter"/> class.
+        /// And allocates the filter natively. 
+        /// <param name="name">The unique name with which this filtered is registered with</param>
+        /// <param name="attributes">Either a single attribute, or a comma separated list of filterForAttributes for which this filter applies to</param>
+        /// </summary>
+        private Filter(string name, string attributes)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(name, "name");
+            Ensure.ArgumentNotNullOrEmptyEnumerable(attributes, "filterForAttributes");
```
jamill: s/filterForAttributes/attributes ?

LibGit2Sharp/Filter.cs
```diff
@@ -0,0 +1,253 @@
...
+        /// <summary>
+        /// The filter filterForAttributes.
+        /// </summary>
+        public IEnumerable<string> Attributes
+        {
+            get { return attributes.Split(','); }
+        }
+
+        /// <summary>
+        /// The marshelled filter
```
jamill: marshalled?

LibGit2Sharp/Filter.cs
```diff
@@ -0,0 +1,253 @@
...
+        /// Decides if a given source needs to be filtered by checking if the filter
+        /// matches the current file extension.
+        /// </summary>
+        /// <param name="filterForAttributes">The filterForAttributes that this filter was created for.</param>
+        /// <param name="filterSource">The source of the filter</param>
+        /// <returns>0 if successful and -30 to skip and pass through</returns>
+        protected virtual int Check(IEnumerable<string> filterForAttributes, FilterSource filterSource)
+        {
+            var fileInfo = new FileInfo(filterSource.Path);
+            var matches = filterForAttributes.Any(currentExtension => fileInfo.Extension == currentExtension);
```
jamill: I think we should use an explicit StringComparison function (`string.Equals(x, y, StringComparison)`)

LibGit2Sharp/Filter.cs
```diff
@@ -0,0 +1,253 @@
...
+            return matches ? 0 : -30;
```
jamill: maybe used a named constant for -30?

LibGit2Sharp/Core/Handles/GitFilterSafeHandle.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Runtime.InteropServices;
+
+namespace LibGit2Sharp.Core.Handles
+{
+    internal class GitFilterSafeHandle : NotOwnedSafeHandleBase
+    {
+        internal GitFilterSafeHandle(GitFilter gitFilter)
```
jamill: I am not sure I understand why we need this constructor (not saying there isn't a good reason...). I would expect this structure to be marshaled correctly to the native layer without being converted into a `GitFilterSafeHandle`. Also, we are allocating memory on the heap with the `Marshal.AllocHGlobal`, but I don't see where this is ever freed. I think this will leak memory.

LibGit2Sharp/Core/Handles/GitFilterSafeHandle.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Runtime.InteropServices;
+
+namespace LibGit2Sharp.Core.Handles
+{
+    internal class GitFilterSafeHandle : NotOwnedSafeHandleBase
```
jamill: Unless there is something I am missing (quite possible), I think I would remove the `GitFilterSafeHandle` class, and just marshal the `GitFilter` type to / from the native layer

LibGit2Sharp/Core/Proxy.cs
```diff
@@ -807,6 +807,50 @@ public static GitDiffDelta git_diff_get_delta(DiffSafeHandle diff, int idx)
...
+            using (ThreadAffinity())
+            {
+                int res = NativeMethods.git_filter_unregister(name);
+                Ensure.ZeroResult(res);
+            }
+        }
+
+        public static FilterMode git_filter_source_mode(IntPtr filterSource)
+        {
+            using (ThreadAffinity())
```
jamill: I think we only **have to** use the ThreadAffinity pattern when we check the error code from libgit (which might require LibGit2Sharp going back to the thread local storage to get the error message set by libgit2).

@nulltoken - My understanding is that the ThreadAffinity call is not actually functionally useful, or even used, by the library, but it causes a lot of confusion about when it should actually be used. What would you think about just removing these?

LibGit2Sharp/Core/GitFilter.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        public string ManagedAttributes()
+        {
+            return GetAttributesFromPointer(attributes);
+        }
+
+        public static IntPtr GetAttributesFromManaged(string attributes)
+        {
+            return EncodingMarshaler.FromManaged(Encoding.UTF8, attributes);
```
jamill: Do we need to free the memory allocated on the heap by `EncodingMarshaler.FromManaged`?

LibGit2Sharp/Core/NativeMethods.cs
```diff
@@ -166,6 +166,9 @@ private static bool IsRunningOnLinux()
         internal static extern IntPtr git_blob_rawcontent(GitObjectSafeHandle blob);
 
         [DllImport(libgit2)]
+        internal static extern IntPtr git_blob_rawcontent(IntPtr blob);
```
jamill: Is this change necessary? I didn't see it used anywhere...

LibGit2Sharp/Core/Proxy.cs
nulltoken: >I think we only have to use the ThreadAffinity pattern when we check the error code from libgit 

Indeed.

>My understanding is that the ThreadAffinity call is not actually functionally useful, or even used, by the library, but it causes a lot of confusion about when it should actually be used. What would you think about just removing these?

@jamill Why not? I've created https://github.com/libgit2/libgit2sharp/issues/953 to discuss about this.

LibGit2Sharp/Core/GitFilter.cs
nulltoken: We should

LibGit2Sharp/GlobalSettings.cs
nulltoken: If that makes sense to keep for that long a hold on a string, couldn't we rather keep hold on to an instance of a type?

Besides, considering @dahlbyk's idea in https://github.com/libgit2/libgit2sharp/pull/948#discussion_r24564786, we may very soon end up returning a type.

LibGit2Sharp/GlobalSettings.cs
```diff
@@ -108,5 +109,46 @@ public static LogConfiguration LogConfiguration
...
+        }
+
+        /// <summary>
+        /// Remove the filter from the registry, and frees the native heap allocation.
+        /// </summary>
+        public static void DeregisterFilter(Filter filter)
+        {
+            DeregisterFilter(filter.Name);
+
+            var disposableFilter = filter as IDisposable;
```
jamill: I am not sure I understand the intended lifecycle of these filters... is it OK to deregister and register the same filter? Is it OK that deregistering a filter by name has different side effects than deregistering a filter through this method? I don't think LibGit2Sharp is the one constructing these objects, maybe we should let the caller manage the lifetime.

dahlbyk: > I don't think LibGit2Sharp is the one constructing these objects, maybe we should let the caller manage the lifetime.

I agree with this.

nulltoken: >> I don't think LibGit2Sharp is the one constructing these objects, maybe we should let the caller manage the lifetime.

>I agree with this.

I also agree with this. My fault. Sorry @ammeep I led you on a wrong track. :confused:

LibGit2Sharp/GlobalSettings.cs
nulltoken: Or, even better (from an Api perspective), I'd :+1: mimicking what @ethomson did with the `SmartSubTransports`. When registering one, the caller gets a `SmartSubTransportsRegistration` that is required for deregistering.

LibGit2Sharp/Core/Handles/GitFilterSafeHandle.cs
ammeep: I could also be missing something here, but without the heap allocation prior to registering the filer I am unable to successfully have any callback made to my filter. Exceptions are thrown due to corrupt memory on when a file is stage (this is when the check/apply callbacks are made)

```

failed: System.AccessViolationException : Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
	at LibGit2Sharp.Core.NativeMethods.git_index_add_bypath(IndexSafeHandle index, FilePath path)
```

I'm about 90% sure that when a filter is deregistered in libgit2 that this memory is free at that point. But would :heart: someone to verify this https://github.com/libgit2/libgit2/blob/d24a5312d8ab6d3cdb259e450ec9f1e2e6f3399d/src/filter.c#L299

If that is correct, then we should be ok to have ``GitFilterSafeHandle`` inherit from ``NotOwnedSafeHandleBase`` because libgit2 handles the lifetime?

LibGit2Sharp/GitBufReader.cs
```diff
@@ -0,0 +1,72 @@
+using System;
+using System.Globalization;
+using System.IO;
+using LibGit2Sharp.Core;
+using LibGit2Sharp.Core.Handles;
+
+namespace LibGit2Sharp
```
nulltoken: Cheating.... :stuck_out_tongue_winking_eye:

ammeep: lol. 

I think I might wait until your branches have been merged to fix these build failures up properly. That sound ok?

LibGit2Sharp/GitBufReader.cs
```diff
@@ -0,0 +1,72 @@
...
+                    string.Format(
+                        CultureInfo.InvariantCulture,
+                        "Provided length ({0}) exceeds long.MaxValue ({1}).",
+                        len.ToUInt64(), long.MaxValue));
+            }
+
+            return (long)len.ToUInt64();
+        }
+
+        private unsafe static Stream WrapGifBufPointer(IntPtr gitBufPointer)
```
jamill: WrapGitBufPointer?

LibGit2Sharp/GlobalSettings.cs
ammeep: I can't seem to find this ``SmartSubTransportsRegistration`` you speak of. Do you know where I can look? #blind :eyes:

LibGit2Sharp/Core/GitBufWriteStream.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            Proxy.git_buf_put(gitBufPointer, buffer, offset, count);
+        }
+
+        private void AutoGrowBuffer(int count)
+        {
+            var gitBuf = gitBufPointer.MarshalAs<GitBuf>();
+
+            var asize = (uint)gitBuf.asize;
+            var size = (uint)gitBuf.size;
```
nulltoken: @ammep THE @ethomson discretly slapped me behind the head. Both `size` and `asize`  should be `ulong`. :blush: Sorry

LibGit2Sharp/Core/GitBufWriteStream.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            var isBufferLargeEnoughToHoldTheNewData = (asize - size) > count;
+            var filledBufferPercentage = (100.0 * size / asize);
+
+            if (isBufferLargeEnoughToHoldTheNewData && filledBufferPercentage < 90)
+            {
+                return;
+            }
+
+            var targetSize = (uint)(1.5 * (asize + count));
```
nulltoken: Similarly, `ulong` here as well... My fault... again

ammeep:
Ok, I think this is all the outstanding feedback addressed. Would love some more :eyes: on this

LibGit2Sharp.Tests/FilterFixture.cs
```diff
@@ -0,0 +1,554 @@
...
+using System.Linq;
+using LibGit2Sharp.Core;
+using LibGit2Sharp.Tests.TestHelpers;
+using Xunit;
+
+namespace LibGit2Sharp.Tests
+{
+    public class FilterFixture : BaseFixture
+    {
+        private const int GitPassThrough = -30;
```
jamill: I think we could use the `GitErrorCode.PassThrough` in the test fixtures. This might help people looking for code samples to not hard code `-30`

LibGit2Sharp/Filter.cs
```diff
@@ -0,0 +1,247 @@
...
+            return 0;
+        }
+
+        /// <summary>
+        /// Decides if a given source needs to be filtered by checking if the filter
+        /// matches the current file extension.
+        /// </summary>
+        /// <param name="filterForAttributes">The filterForAttributes that this filter was created for.</param>
+        /// <param name="filterSource">The source of the filter</param>
+        /// <returns>0 if successful and -30 to skip and pass through</returns>
```
jamill: Maybe the comments could refer to `GitErrorCode.PassThrough` instead of an actual number?

LibGit2Sharp/FilterRegistration.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        internal IntPtr FilterPointer { get; private set; }
+
+        internal void Free()
+        {
+            Marshal.FreeHGlobal(FilterPointer);
+            FilterPointer = IntPtr.Zero;
+        }
+    }
+}
```
nulltoken: Missing trailing newline :trolleybus:

LibGit2Sharp/FilterSource.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// The relative path to the file
+        /// </summary>
+        public virtual string Path { get; private set; }
+
+        /// <summary>
+        /// The blob id
+        /// </summary>
+        public virtual ObjectId ObjectId { get; private set; }
+    }
+}
```
nulltoken: Here as well

LibGit2Sharp/Filter.cs
```diff
@@ -0,0 +1,247 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using LibGit2Sharp.Core;
+
+namespace LibGit2Sharp
+{
+    /// <summary>
+    /// A filter
```
jamill: We could include more text about what this filter class is used (what does this filter?)

LibGit2Sharp/Filter.cs
```diff
@@ -0,0 +1,247 @@
...
+    {
+        private static readonly LambdaEqualityHelper<Filter> equalityHelper =
+        new LambdaEqualityHelper<Filter>(x => x.Name, x => x.Attributes);
+
+        private readonly string name;
+        private readonly string attributes;
+
+        private readonly GitFilter managedFilter;
+
+
```
jamill: nit: extra line :)

nulltoken:
Looks clean to me. I'll take a (fresher) look tomorrow.

We'd still need https://github.com/libgit2/libgit2/issues/2892 to be merged (and update the libgit2 binaries) before pulling this in .

Very awesome job for someone with a *"rusty knowledge of C"* who stated *"this is the first time I have looked at libgit2sharp in any great detail"* :stuck_out_tongue_winking_eye:

LibGit2Sharp/FilterMode.cs
```diff
@@ -0,0 +1,20 @@
...
+        /// </summary>
+        Smudge = 0,
+
+        /// <summary>
+        /// Clean occurs when importing a file from the working directory to the Git object database.
+        /// For example, a file would be cleaned when staging a file.
+        /// </summary>
+        Clean = (1 << 0),
+    }
+}
```
jamill: nit: trailing newline

LibGit2Sharp/Filter.cs
```diff
@@ -0,0 +1,247 @@
...
+        /// </summary>
+        public IEnumerable<string> Attributes
+        {
+            get { return attributes.Split(','); }
+        }
+
+        /// <summary>
+        /// The marshalled filter
+        /// </summary>
+        internal GitFilter ManagedFilter
```
jamill: nit: `ManagedFilter` is a bit confusing to me - I think I would tweak to something that indicates this returns the struct to be marshalled to the native layer (as your comment alludes to).

LibGit2Sharp/GlobalSettings.cs
```diff
@@ -108,5 +109,38 @@ public static LogConfiguration LogConfiguration
                 return logConfiguration;
             }
         }
+
+        /// <summary>
+        /// Register a filter globally with a default priority of 200 allowing the custom filter 
+        /// to imitate a core Git filter driver. It will be run last on checkout and first on checkin.
+        /// </summary>
+        public static FilterRegistration RegisterFilter(Filter filter)
```
jamill: Just a thought:

The only public property of `FilterRegistration` is the name. Instead of making the consumer responsible for keeping `FilterRegistration` objects around, what if we bookkept them internally (and make these objects internal only). We could then just return the filtername during registration, allow them to deregister by name, and possible even allow them to query known filter registrations. This way, we would not hand out objects that could be "lost" by the consumers (that has heap allocated memory)

nulltoken: I'd rather stick on the same pattern we've set up with the `SmartSubTransport` registrations. That would reduce the chance for a user to unregister a filter that he has been registered by another user operating in the same AppDomain.

>This way, we would not hand out objects that could be "lost" by the consumers (that has heap allocated memory)

How about making them both implement IDisposable and a finalizer? (This could be tackled in a separate PR)

jamill: My proposal has the exact same semantics regarding the filter registration / deregistration as exists now - they would still only be able to deregister what was registered in their app domain. It is just that Libgit2Sharp would track the registrations.

Sent from my Windows Phone
________________________________

dahlbyk: How about both:

1. Bookkeep internally to facilitate deregistration by name. We'd need to decide if it's an error or noop to deregister a filter that wasn't registered by this app domain. Maybe return a `bool` indicating successful `DeregisterFilter()`?
2. Continue to return a `FilterRegistration` that implements `IDisposable` with a finalizer because unmanaged, and also because `using (GlobalSettings.RegisterFilter(myFilter)) { ... }` still strikes me as a useful API. For scenarios where retaining the `FilterRegistration` is inconvenient, we'd still offer deregistration by name.

nulltoken: For now, I'd rather stick on the same pattern we've set up with the `SmartSubTransport` registrations and avoid deregistration through the filter name.

Once this is merged, let's create a separate issue to discuss this topic further (unmanaged memory reclaiming on GC, deregistration process, ...). FWIW, if we made `xxxRegistration` types disposables, we may even get rid of the `Deregister` methods...

nulltoken:
See https://github.com/libgit2/libgit2sharp/pull/958 for a potential improvement proposal

LibGit2Sharp/Core/GitBufWriteStream.cs
```diff
@@ -0,0 +1,92 @@
...
+    internal class GitBufWriteStream : Stream
+    {
+        private readonly IntPtr gitBufPointer;
+
+        internal GitBufWriteStream(IntPtr gitBufPointer)
+        {
+            this.gitBufPointer = gitBufPointer;
+
+            //Preallocate the buffer
+            Proxy.git_buf_grow(gitBufPointer, 1024);
```
nulltoken: As pointed out by @carlosmn , let's make this `4096`

nulltoken:
@ammeep I've added a tiny commit to get the feedback from AppVeyor when time's ready.

When updating libgit2 binaries, we tend to stick to a convention that ease further troubleshooting and use this kind of message

```
Update libgit2 to fb2f3a7

https://github.com/libgit2/libgit2/compare/e0902fb...fb2f3a7
```

ammeep:
Ooops, sorry @nulltoken I'll fix that up :soon:

Right now I'm trying to figure out the support for our :sparkles: new streaming bindings. Once that's one I'll rebase the whole thing again :+1:

nulltoken:
@ammeep No need to be sorry! It's an (undocumented) convention :wink:

Thanks for this amazing work!

nulltoken:
IT'S IN! @ammeep All you wonderful hard work has made it in #1030! :heart_eyes_cat:

PR #915, opened 2015-01-16
Wrap git_trace_set to hold a reference to trace method to prevent GC.

jeffhostetler:
Have NativeMethod layer hold a reference to the trace handler that we give to LibGit2 so that it won't be GC'd if the caller doesn't persist it properly.

I do question why this was necessary since GlobalSettings is holding a static LogConfiguration which contains a reference to the handler.

I don't have a test for the change since I couldn't figure out how to cause logging to happen from a test since everything is private and libgit2 doesn't have many calls to logging.

Please advise.
Thanks.

### DIFF
--- a/LibGit2Sharp/Core/NativeMethods.cs
+++ b/LibGit2Sharp/Core/NativeMethods.cs
@@ -1503,8 +1503,30 @@ namespace LibGit2Sharp.Core
 
         internal delegate void git_trace_cb(LogLevel level, IntPtr message);
 
-        [DllImport(libgit2)]
-        internal static extern int git_trace_set(LogLevel level, git_trace_cb trace_cb);
+        /// <summary>
+        /// A variable to hold a reference to the installed callback function
+        /// because it needs to persist until another callback is installed
+        /// or tracing is disabled.  The trace callback is special because its
+        /// use is not limited to the duration of the downcall.
+        /// </summary>
+        private static git_trace_cb static_trace_cb;
+
+        /// <summary>
+        /// Make git_trace_set() a wrapper around actual library routine to
+        /// ensure that we hold a reference to the callback.
+        /// </summary>
+        public static int git_trace_set(LogLevel level, git_trace_cb trace_cb)
+        {
+            static_trace_cb = trace_cb;
+            return _dll__git_trace_set(level, static_trace_cb);
+        }
+
+        /// <summary>
+        /// Forward calls to _dll__git_trace_set() to the actual git_trace_set()
+        /// inside LibGit2.dll
+        /// </summary>
+        [DllImport(libgit2, EntryPoint="git_trace_set")]
+        internal static extern int _dll__git_trace_set(LogLevel level, git_trace_cb trace_cb);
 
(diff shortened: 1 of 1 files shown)


LibGit2Sharp/Core/NativeMethods.cs
```diff
@@ -1503,8 +1503,30 @@ private static bool IsRunningOnLinux()
...
+        /// </summary>
+        private static git_trace_cb static_trace_cb;
+
+        /// <summary>
+        /// Make git_trace_set() a wrapper around actual library routine to
+        /// ensure that we hold a reference to the callback.
+        /// </summary>
+        public static int git_trace_set(LogLevel level, git_trace_cb trace_cb)
+        {
+            static_trace_cb = trace_cb;
```
Therzok: Why not GC.KeepAlive in the caller?

jeffhostetler: I suppose, but that means all callers of git_trace_set() would need to do that to be completely safe.  There's also the issue of when/if those callers could/should undo the KeepAlive on the method.

To me, it seemed that since this routine is stuffing a handle/object into LibGit2 for later use, it is responsible for holding the reference.  And that reference should persist until we either clear or pass down a different object.  (I guess technically, LibGit2 should call back into the CLR to declare that it is holding the reference, but this seemed more straight-forward.)

Therzok: KeepAlive is just a an empty non-inlined method that will keep the object alive. What 'undo' are you talking about? :P

If this should persist between sessions, I'd rather have this documented and tell the user. /cc @nulltoken 

Having it kept inside LibGit2Sharp would mean never having it freed. That can impact the user in case we're using delegates which keep other things in memory. Rather leave it in user control rather than a cache that might fail way harder than we think it should?

Therzok: For GC.KeepAlive check [this](http://msdn.microsoft.com/en-us/library/system.gc.keepalive%28v=vs.110%29.aspx) and see how it works. It just makes an empty call that will trick the GC into thinking that the variable is used, so it can't reclaim it.

jamill:
Thanks @jeffhostetler for this!

I am not sure if using GC.KeepAlive would help here - I think the problem is that there is no managed reference to the delegate past the scope of the `Proxy.git_trace_set` function, so calling `GC.KeepAlive` at the end of the function would not be sufficient.

I wonder if explicitly storing the delegate for `GitTraceHandler` in `LogConfiguration`, and passing that delegate  instead of the `GitTraceHandler` to `Proxy.git_trace_set` would also prevent the garbage collection issue.

Assuming the above proposal even works, I think either the current solution proposed by @jeffhostetler (but I would move the static variable logic into the `Proxy` layer), or putting the onus on callers into `Proxy.git_trace_set` and documenting the method as @Therzok suggested would work. Keeping the logic closer to the native function call would make it harder for callers to misuse this function in a subtle way, which is nice. We control all the callers into `Proxy.git_trace_set` and how they set up the callback.

jeffhostetler:
Thanks to both @jamill  and @Therzok for the suggestions and comments.  I found a better solution and will close this PR and submit a new PR with that change.

PR #929, opened 2015-01-23
Expose the assume-unchanged flag to IndexEntry #928

zpete:
Didn't know where to put the constant GIT_IDXENTRY_VALID to - I hope LibGit2Sharp/Core/GitIndexEntry.cs is the right place.
This is my first Pull request, so bear with me.


LibGit2Sharp/IndexEntry.cs
```diff
@@ -31,6 +31,11 @@ public class IndexEntry : IEquatable<IndexEntry>
         public virtual StageLevel StageLevel { get; private set; }
 
         /// <summary>
+        /// Whether the file is marked as assume-unchanged
+        /// </summary>
+        public bool AssumeUnchanged { get; private set; }
```
nulltoken: Please mark this property as `virtual`. This should allow the build to pass :wink:

nulltoken:
Beside the tiny nitpick above, I :heart: this very much :+1: 

One thing would be even awesomer: While amending your commit, could you please reword its commit message so that the reference to the issue is on the third line? Hint: If you write `Fix #928` instead of `Issue #928`, the initial issue will be automatically closed once this commit is merged.

Something like this maybe

```
Expose the assume-unchanged flag to IndexEntry

Fix #928 
```

Another thing. You may be willing to update your GitHub profile and add the email address you've used in this commit as a known one. This way, your contributions will be linked to your GitHub account.

nulltoken:
:cool:

One last request.  As the second commit is only a fixup of the first one, would you please squash them together?

zpete:
Thank you very much for all the useful help. I hope everything is fine with
the Pull request now. Let me know if something is still wrong ;-)

carlosmn:
:+1: as we get more flags, we'd eventually want to create its own type, like we have for the other flags, but for one it's fine as it is.

nulltoken:
:+1: for me as well.

nulltoken:
@zpete :sparkles: :sparkles: :sparkles: :sparkles: :sparkles: :sparkles: :sparkles: :bangbang:

PR #938, opened 2015-01-30
add support for testing merge conflicts

### LINKED ISSUE #628: Test a branch can merge or not
In order to make a opened Git platform like the feature Pull-Request, I proposal introduce a instance method `bool Repository.CanMerge(Commit first, Commit second)` that return true if `second` can merge into `first`, otherwise return false.

shiftkey:
Fixes #628

- ~~[ ] helper functions to simplify discovering trees?~~ just use `Commit`
- [x] other scenarios to test?
- [x] correct rebase to drop `Repository.Version`
- [x] clean up history

A couple of questions for reviewers:

 - ~~we kept the API as low-level as possible - should we add overloads to lookup by a `commitish` string?~~ switched over to use `Commit` instead of `Tree`
 - ~~the `bool` returned from the API, is that enough detail?~~ perhaps a more detailed API in the future
 - i tested the basic scenarios of "self merge", "clean merge" and "merge with conflicts" - can you think of any others we should include here?

### DIFF
--- a/LibGit2Sharp/ObjectDatabase.cs
+++ b/LibGit2Sharp/ObjectDatabase.cs
@@ -406,5 +406,34 @@ namespace LibGit2Sharp
 
             return gitObject.Sha.Substring(0, minLength.Value);
         }
+
+        /// <summary>
+        /// Returns whether merging <paramref name="one"/> into <paramref name="another"/>
+        /// would result in merge conflicts.
+        /// </summary>
+        /// <param name="one">The commit wrapping the base tree to merge into.</param>
+        /// <param name="another">The commit wrapping the tree to merge into <paramref name="one"/>.</param>
+        /// <returns>True if the merge does not result in a conflict, false otherwise.</returns>
+        public virtual bool CanMergeWithoutConflict(Commit one, Commit another)
+        {
+            Ensure.ArgumentNotNull(one, "one");
+            Ensure.ArgumentNotNull(another, "another");
+
+            using (var ourHandle = Proxy.git_object_peel(repo.Handle, one.Id, GitObjectType.Tree, true))
+            using (var theirHandle = Proxy.git_object_peel(repo.Handle, another.Id, GitObjectType.Tree, true))
+            {
+                var ancestorCommit = repo.Commits.FindMergeBase(one, another);
+
+                var ancestorHandle = ancestorCommit != null
+                    ? Proxy.git_object_peel(repo.Handle, ancestorCommit.Id, GitObjectType.Tree, false)
+                    : new NullGitObjectSafeHandle();
+
+                using (ancestorHandle)
+                using (var indexHandle = Proxy.git_merge_trees(repo.Handle, ancestorHandle, ourHandle, theirHandle))
+                {
+                    return !Proxy.git_index_has_conflicts(indexHandle);
+                }
+            }
+        }
     }
 }
--- a/LibGit2Sharp.Tests/ObjectDatabaseFixture.cs
+++ b/LibGit2Sharp.Tests/ObjectDatabaseFixture.cs
@@ -654,6 +654,50 @@ namespace LibGit2Sharp.Tests
             }
         }
 
+        [Fact]
+        public void TestMergeIntoSelfHasNoConflicts()
+        {
+            string path = SandboxMergeTestRepo();
+            using (var repo = new Repository(path))
+            {
+                var master = repo.Lookup<Commit>("master");
+
+                var result = repo.ObjectDatabase.CanMergeWithoutConflict(master, master);
+
+                Assert.True(result);
+            }
+        }
+
+        [Fact]
+        public void TestMergeIntoOtherBranchHasNoConflicts()
+        {
+            string path = SandboxMergeTestRepo();
+            using (var repo = new Repository(path))
+            {
+                var master = repo.Lookup<Commit>("master");
+                var branch = repo.Lookup<Commit>("fast_forward");
+
+                var result = repo.ObjectDatabase.CanMergeWithoutConflict(master, branch);
+
+                Assert.True(result);
+            }
+        }
+
+        [Fact]
+        public void TestMergeIntoWrongBranchHasConflicts()
+        {
+            string path = SandboxMergeTestRepo();
+            using (var repo = new Repository(path))
+            {
+                var master = repo.Lookup<Commit>("master");
+                var branch = repo.Lookup<Commit>("conflicts");
+
+                var result = repo.ObjectDatabase.CanMergeWithoutConflict(master, branch);
+
+                Assert.False(result);
+            }
+        }
+
         private static Blob CreateBlob(Repository repo, string content)
         {
             using (var stream = new MemoryStream(Encoding.UTF8.GetBytes(content)))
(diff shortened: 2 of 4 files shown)


ethomson:
Does this method make more sense on `Tree` instead of `Repository`?

Dumb question: is there value in making this more general?  We could juggle some things around and have a `Tree.MergeWith(Tree)` that returns an `Index`, which might be useful.  (But maybe not; there are lifetime rules on an `Index` and exposing an `Index` that isn't the repo's index isn't something we've done here in sharp so there may be some (many?) issues that I haven't thought through.)

ethomson:
The reason I ask is because you asked if `bool` was enough detail.  My initial thought is "no", but I think the level of detail that I want probably is the ability to query everything in the `Index`,

shiftkey:
> Does this method make more sense on `Tree` instead of `Repository`?

I think so.

> Dumb question: is there value in making this more general?  

Not a dumb question at all. The use case which motivated this for us was a need to answer the question "is this even possible to merge?" - the details themselves weren't really relevant to us beyond a :thumbsup:/:thumbsdown:. And as you said, carrying around `Index`es might be fraught with danger.

> but I think the level of detail that I want probably is the ability to query everything in the Index,

Thinking out loud: if you care about the state of the index after a merge, why wouldn't you use the existing merge APIs that are already in place (aside from avoiding the need to checkout the base branch)?

ethomson:
I also don't want to do the merge *on disk*, but I do want to be able to speak more intelligently about the conflicts than "yes" or "no".  ("File `X` conflicted because it was changed in `branch` and renamed in `master`, etc.)

(This is what we do now speaking to libgit2, I doubt that we're in any hurry to make this .NET code, but I suspect we will some day...)

shiftkey:
What about if we had two operations for this, just to test the waters:

 - `bool Tree.MergeHasConflicts(Tree ours, Tree theirs)` - this is the one we implement now to get a simple yes/no answer
 - `Index Tree.Merge(Tree ours, Tree theirs)` - this is the one we could implement nor or later which gives you more details.

I'm kinda pushing back here because I don't want to implement APIs without a big demand that would cause the maintainers headaches, but this compromise could let both live side-by-side...

nulltoken:
>> Does this method make more sense on `Tree` instead of `Repository`?

>I think so.

I'd prefer to keep the `GitObject`s as dumb as possible. How would you hate making `ObjectDatabase` expose it? 

Going a bit further, how about making this simpler for the caller with the following?

```
bool ObjectDatabase.CanMergeWithoutConflict(Commit ours, Commit theirs)
```

> but I think the level of detail that I want probably is the ability to query everything in the Index,

@ethomson Are you only after querying capabilities or would you also require to update it? I feel a little bit wary about returning the `Index` type as the caller may see it as *the* Index rather than *an* Index. Maybe should we return a different type that would reflect the in-memory only aspect of it? 

>I'm kinda pushing back here because I don't want to implement APIs without a big demand

I *think* I see some of the pros of this API. However, I'd :+1: to deal with it in another PR/Issue as it may require some more time to finalize it.

LibGit2Sharp/Repository.cs
```diff
@@ -981,6 +981,44 @@ private void CleanupDisposableDependencies()
...
 
         /// <summary>
+        /// Gets the current LibGit2Sharp version.
+        /// <para>
+        ///   The format of the version number is as follows:
+        ///   <para>Major.Minor.Patch-LibGit2Sharp_abbrev_hash-libgit2_abbrev_hash (x86|amd64 - features)</para>
+        /// </para>
+        /// </summary>
+        [Obsolete("This property will be removed in the next release. Use GlobalSettings.Version instead.")]
+        public static string Version
```
nulltoken: Isn't this already dropped from vNext?

shiftkey: You're right, just a bad rebase on my part

shiftkey:
@nulltoken 

>  How would you hate making ObjectDatabase expose it?

I wouldn't hate it at all :grinning: 

> Going a bit further, how about making this simpler for the caller with the following? ...

I can totally do that.

> However, I'd :+1: to deal with it in another PR/Issue as it may require some more time to finalize it.

I'll move ahead with just this API in the PR, but feel free to keep putting our thoughts around other things here - I'll move the net result into a new issue when this is done.

ethomson:
> @ethomson Are you only after querying capabilities or would you also require to update it? I feel a little bit wary about returning the Index type as the caller may see it as the Index rather than an Index. Maybe should we return a different type that would reflect the in-memory only aspect of it?

I am only querying the resultant `Index` so that I can determine if it is mergeable, and if not, to produce the list of files that are in conflict.  I do not need to write to it.  It could, indeed, be some other type.

shiftkey:
Ok, so that's been updated to use `ObjectDatabase` (and the tests have been migrated). 

Let me know how you feel about it in it's current form.

LibGit2Sharp/ObjectDatabase.cs
```diff
@@ -406,5 +406,31 @@ public virtual string ShortenObjectId(GitObject gitObject, int? minLength = null
 
             return gitObject.Sha.Substring(0, minLength.Value);
         }
+
+        /// <summary>
+        /// Returns whether merging <paramref name="ours"/> into <paramref name="theirs"/>
+        /// would result in merge conflicts.
+        /// </summary>
+        /// <param name="ours">The base tree to merge into.</param>
```
nulltoken: Maybe `the commit wrapping the base tree to merge into`?

LibGit2Sharp/ObjectDatabase.cs
```diff
@@ -406,5 +406,31 @@ public virtual string ShortenObjectId(GitObject gitObject, int? minLength = null
...
+        /// <param name="theirs">The tree to merge into <paramref name="ours"/>.</param>
```
nulltoken: Similarly here

nulltoken:
@Aimeast How do you feel about this?

nulltoken:
@shiftkey Could we simplify the history, by maybe melding "cleanup that icky conditional" and "moved method into ObjectDatabase" into the original commit? I can take of this while merging this PR, if you want.

LibGit2Sharp/ObjectDatabase.cs
nulltoken: Thinking a bit more, considering the use case of returning a bool, is there a real "direction" (into, ours, theirs, ...).

Couldn't we be simpler to understand for the reader? Maybe just by naming the parameters "one" and "another"? Unless I'm wrong, we can safely swap the values without changing the output of the method.

shiftkey:
@nulltoken sure thing, I'll add a task to squash this history together so it doesn't get missed

nulltoken:
:heart_eyes_cat:

LibGit2Sharp/ObjectDatabase.cs
shiftkey: > Unless I'm wrong, we can safely swap the values without changing the output of the method.

That's a good point. I'm not sure there's a situation where it should work one way and not the other. I do like making them *similar* to the underlying `libgit2` parameters but I'm not attached to these names.

> Couldn't we be simpler to understand for the reader?

`one` and `another` sounds good. Another pair I had in mind was `@base` (ew, reserved) and `commit`, but I'll try yours on first.

shiftkey:
:8ball: I think that's all the feedback addressed

nulltoken:
@joshaber :sparkling_heart: :bangbang:

nulltoken:
@joshvera @shiftkey sparkling_heart: :bangbang:

nulltoken:
Published in prerelease NuGet package `LibGit2Sharp.0.22.0-pre20150207140248.nupkg`

nulltoken:
yeah_59002_gifsoup com

PR #968, opened 2015-02-20
No more check function

ammeep:
This PR removes the need to supply the check callback when registering a Filter with the library.

For example if you want to install a filter with libgit2 with the name ``foo`` you would register that filter with libgit2sharp and it would be applied to any file that matched the path spec found in the gitattributes file. For example `` *.txt filter=foo``

I naively thought that we needed to implement custom check logic for filters (the piece that actually matches the entry in the attributes file to the file being filtered) but it turns out this is already implemented in libgit2 https://github.com/libgit2/libgit2/blob/master/src/filter.c#L403 and if we simply do not provide the optional check callback, everything just works :tm:

I've added an object ``FilterAttribute`` which allows you to define and register a filter in the correct format ``filter=name``

### DIFF
--- a/LibGit2Sharp/Filter.cs
+++ b/LibGit2Sharp/Filter.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text;
 using LibGit2Sharp.Core;
 
@@ -16,7 +17,7 @@ namespace LibGit2Sharp
         new LambdaEqualityHelper<Filter>(x => x.Name, x => x.Attributes);
 
         private readonly string name;
-        private readonly string attributes;
+        private readonly IEnumerable<FilterAttribute> attributes;
 
         private readonly GitFilter gitFilter;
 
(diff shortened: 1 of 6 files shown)


ammeep:
I would love some eyes on this @nulltoken. 

Want to make sure it is sane before I merge it, squash down the register filter PR, and update libgit2 to the latest :nail_care:

LibGit2Sharp/Filter.cs
```diff
@@ -25,32 +25,22 @@ public abstract class Filter : IEquatable<Filter>
...
-        private Filter(string name, string attributes)
+        protected Filter(string name, IEnumerable<FilterAttribute> attributes)
         {
             Ensure.ArgumentNotNullOrEmptyString(name, "name");
-            Ensure.ArgumentNotNullOrEmptyEnumerable(attributes, "attributes");
+            Ensure.ArgumentNotNull(attributes, "attributes");
 
             this.name = name;
             this.attributes = attributes;
+            var attributesAsString = string.Join(",", this.attributes.Select(attr => attr.FilterDefinition));
```
Therzok: Nit: Just use the linq inside the initializer list?

LibGit2Sharp/FilterAttribute.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        private readonly string filterDefinition;
+
+        /// <summary>
+        /// The name of the filter found in a .gitattributes file
+        /// </summary>
+        /// <param name="filterName">The name of the filter</param>
+        public FilterAttribute(string filterName)
+        {
+            Ensure.ArgumentNotNullOrEmptyString(filterName, "filterName");
```
Therzok: Shouldn't this check if it's null or whitespace?

ammeep: Maybe this method needs a rename, but I don't think I should do this here :+1: 
https://github.com/libgit2/libgit2sharp/blob/vNext/LibGit2Sharp/Core/Ensure.cs#L53

Therzok: Ah, then it's fine. Bad naming.

PR #1004, opened 2015-03-18
Duplicate Filepaths in a commit causes failure when performing Diff between commit tree and its parent

nulltoken:
Duplicate FilePaths in a commit causes failure when trying to perform Diff between commit tree and its parent
This happens in TreeChanges.cs class in AddChanges method, as Dictionary prevents addition of same key.

### DIFF
--- a/LibGit2Sharp.Tests/DiffTreeToTreeFixture.cs
+++ b/LibGit2Sharp.Tests/DiffTreeToTreeFixture.cs
@@ -39,7 +39,7 @@ namespace LibGit2Sharp.Tests
 
                 var changes = repo.Diff.Compare<TreeChanges>(tree, tree);
 
-                Assert.Null(changes["batman"]);
+                Assert.Equal(0, changes.Count(c => c.Path == "batman"));
             }
         }
 
@@ -62,7 +62,7 @@ namespace LibGit2Sharp.Tests
                 Assert.Equal(1, changes.Count());
                 Assert.Equal(1, changes.Added.Count());
 
-                TreeEntryChanges treeEntryChanges = changes["1.txt"];
+                TreeEntryChanges treeEntryChanges = changes.Single(c => c.Path == "1.txt");
 
                 var patch = repo.Diff.Compare<Patch>(parentCommitTree, commitTree);
                 Assert.False(patch["1.txt"].IsBinaryComparison);
@@ -272,8 +272,7 @@ namespace LibGit2Sharp.Tests
                 var changes = repo.Diff.Compare<TreeChanges>(rootCommitTree, commitTreeWithRenamedFile);
 
                 Assert.Equal(1, changes.Count());
-                Assert.Equal("super-file.txt", changes["super-file.txt"].Path);
-                Assert.Equal("my-name-does-not-feel-right.txt", changes["super-file.txt"].OldPath);
+                Assert.Equal("my-name-does-not-feel-right.txt", changes.Single(c => c.Path == "super-file.txt").OldPath);
                 Assert.Equal(1, changes.Renamed.Count());
             }
         }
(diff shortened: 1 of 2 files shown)


nulltoken:
In order to move forward on this topic, I've added the hopefully failing tests. Let's see how the CI servers think of them.

LibGit2Sharp.Tests/DiffTreeToTreeFixture.cs
```diff
@@ -930,45 +929,68 @@ public void CanHandleTwoTreeEntryChangesWithTheSamePath()
...
-                TreeEntryChanges change = changes[path];
-                Assert.Equal(Mode.SymbolicLink, change.OldMode);
-                Assert.Equal(Mode.NonExecutableFile, change.Mode);
-                Assert.Equal(ChangeKind.TypeChanged, change.Status);
-                Assert.Equal(path, change.Path);
-            }
+            CanHandleTwoTreeEntryChangesWithTheSamePath(SimilarityOptions.Default,
+                (path, changes) =>
+                {
+                    // TODO: What should we assert there?
```
nulltoken: @ethomson What should we assert there?

nulltoken:
@SinghVarun @ethomson I've pushed a proposal. A breaking one.

Thoughts?

LibGit2Sharp.Tests/DiffTreeToTreeFixture.cs
```diff
@@ -930,45 +928,61 @@ public void CanHandleTwoTreeEntryChangesWithTheSamePath()
...
+                    Assert.Equal(path, change.Path);
+                });
+        }
+
+        [Fact]
+        public void CanHandleTwoTreeEntryChangesWithTheSamePathUsingSimilarityDefault()
+        {
+            // $ git diff-tree --name-status --find-renames -r 2ccccf8 e829333
+            // T       include/Nu/Nu.h
+            // D       objc/Nu.h
```
nulltoken: This result seems to not match libgit2 output. I may surely have messed my syntax somewhere.

@carlosmn @ethomson Any idea?

LibGit2Sharp.Tests/StatusFixture.cs
```diff
@@ -592,5 +593,57 @@ public void RetrievingTheStatusHonorsAssumedUnchangedMarkedIndexEntries()
...
+
+                var expected = Path.Combine("include", "Nu", "Nu.h");
+
+                foreach (var repositoryStatu in status)
+                {
+                    Trace.WriteLine(string.Format("{0} - {1}", repositoryStatu.FilePath, repositoryStatu.State));
+                }
+
+                if (IsRunningOnUnix())
+                {
```
nulltoken: @carlosmn @ethomson Does this seem legit to you to get different outcomes depending on the platform?

LibGit2Sharp.Tests/StatusFixture.cs
```diff
@@ -592,5 +593,57 @@ public void RetrievingTheStatusHonorsAssumedUnchangedMarkedIndexEntries()
...
+                Touch(parentPath, "Nu.h", "awesome content\n");
+
+                RepositoryStatus status = repo.RetrieveStatus(
+                    new StatusOptions{ DetectRenamesInIndex = true, DetectRenamesInWorkDir = true });
+
+                Assert.Equal(2, status.Count());
+
+                var expected = Path.Combine("include", "Nu", "Nu.h");
+
+                foreach (var repositoryStatu in status)
```
ethomson: `repositoryStatus`?

nulltoken: Leftover Ci side debug code. Will drop this.

LibGit2Sharp.Tests/StatusFixture.cs
ethomson: Probably not.  Which one is the expected output?

nulltoken: Dunno. They both seem to be valid but expressed in different ways.

nulltoken: I'm going to move this test its strange output to a different PR in order keep this one scoped.

nulltoken:
@SinghVarun Any feedback?

SinghVarun:
@nulltoken  Sorry for delay in response. Was testing this fix on multiple repository.  Things look great.
Thanks all, once again for doing the quick fix.

nulltoken:
Published as pre-release NuGet package `LibGit2Sharp.0.22.0-pre20150324065655`

PR #989, opened 2015-03-28
`Index.Count` is wee

nulltoken:
Currently, [`Index.Count`](https://github.com/libgit2/libgit2sharp/blob/a026f5d84e35f1df94bc3f30503429ef8a216209/LibGit2Sharp/Index.cs#L59) returns an `int`.  Should this be a `long`?  (This is a `size_t` in native land.)  Exceeding a signed 32 bit's worth of index entries is a lot, but not an unfathomable or impossible number.

Similarly, [`git_index_reuc_entrycount`](https://github.com/libgit2/libgit2sharp/blob/a026f5d84e35f1df94bc3f30503429ef8a216209/LibGit2Sharp/Core/NativeMethods.cs#L613) and [`git_index_name_entrycount`](https://github.com/libgit2/libgit2sharp/blob/a026f5d84e35f1df94bc3f30503429ef8a216209/LibGit2Sharp/Core/NativeMethods.cs#L591) probably want to be `UIntPtr`s in `NativeMethods` and choked down to a `long` in the `Proxy`.

### DIFF
--- a/LibGit2Sharp/Core/NativeMethods.cs
+++ b/LibGit2Sharp/Core/NativeMethods.cs
@@ -564,7 +564,7 @@ namespace LibGit2Sharp.Core
         internal static extern int git_index_has_conflicts(IndexSafeHandle index);
 
         [DllImport(libgit2)]
-        internal static extern uint git_index_name_entrycount(IndexSafeHandle handle);
+        internal static extern UIntPtr git_index_name_entrycount(IndexSafeHandle handle);
 
         [DllImport(libgit2)]
         internal static extern IndexNameEntrySafeHandle git_index_name_get_byindex(IndexSafeHandle handle, UIntPtr n);
@@ -586,7 +586,7 @@ namespace LibGit2Sharp.Core
 
 
         [DllImport(libgit2)]
-        internal static extern uint git_index_reuc_entrycount(IndexSafeHandle handle);
+        internal static extern UIntPtr git_index_reuc_entrycount(IndexSafeHandle handle);
 
         [DllImport(libgit2)]
         internal static extern IndexReucEntrySafeHandle git_index_reuc_get_byindex(IndexSafeHandle handle, UIntPtr n);
--- a/LibGit2Sharp/Core/Proxy.cs
+++ b/LibGit2Sharp/Core/Proxy.cs
@@ -3461,5 +3461,28 @@ namespace LibGit2Sharp.Core
             return result ? 0 : (int)GitErrorCode.User;
         }
     }
+
+    /// <summary>
+    /// Class to hold extension methods used by the proxy class.
+    /// </summary>
+    static class ProxyExtensions
+    {
+        /// <summary>
+        /// Convert a UIntPtr to a int value. Will throw
+        /// exception if there is an overflow.
+        /// </summary>
+        /// <param name="input"></param>
+        /// <returns></returns>
+        public static int ConvertToInt(this UIntPtr input)
+        {
+            ulong ulongValue = (ulong)input;
+            if (ulongValue > int.MaxValue)
+            {
+                throw new LibGit2SharpException("value exceeds size of an int");
+            }
+
+            return (int)input;
+        }
+    }
 }
 // ReSharper restore InconsistentNaming
--- a/LibGit2Sharp/Core/Proxy.cs
+++ b/LibGit2Sharp/Core/Proxy.cs
@@ -942,12 +942,8 @@ namespace LibGit2Sharp.Core
 
         public static int git_index_entrycount(IndexSafeHandle index)
         {
-            UIntPtr count = NativeMethods.git_index_entrycount(index);
-            if ((long)count > int.MaxValue)
-            {
-                throw new LibGit2SharpException("Index entry count exceeds size of int");
-            }
-            return (int)count;
+            return NativeMethods.git_index_entrycount(index)
+                .ConvertToInt();
         }
 
         public static StageLevel git_index_entry_stage(IndexEntrySafeHandle index)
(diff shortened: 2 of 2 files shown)


nulltoken:
Turned this into an Pull Request

LibGit2Sharp/Index.cs
```diff
@@ -94,10 +94,9 @@ public virtual bool IsFullyMerged
 
         private List<IndexEntry> AllIndexEntries()
         {
-            var entryCount = Count;
-            var list = new List<IndexEntry>(entryCount);
```
nulltoken: Dropped this as capacity parameter is an `int`. It's unclear exactly how this type will behave when being added elements when capacity is around `int.MaxValue`, though.

Thoughts?

nulltoken:
Dammit! Network tests are complex to run when GitHub is under a DDOS attack... :confounded:

LibGit2Sharp/Index.cs
Therzok: https://github.com/mono/referencesource/blob/mono/mscorlib/system/collections/generic/list.cs#L42

If we happen to overflow, oh well. :P

nulltoken:
Better? I eventually got rid of the `List<>`.

Therzok:
Using IEnumerable everywhere is a bit of an overhead, imo. It can bring really bad practices into play™.

ethomson:
@Therzok can you expand on that?

Therzok:
Basically, you have no access to the underlying List implementation when you use `yield return`. So converting to any kind of other collection will be an `O(n)` operation [0] instead of a block copy. So that means slower.

I'd just stick with List or IList here, don't know any benefit of going pure IEnumerable. While we still need to map then and do `O(n)`, I'd rather use a `List` or anything `ICollection` derived and add elements then return it as an IEnumerable.

[0] - https://github.com/Microsoft/referencesource/blob/master/mscorlib/system/collections/generic/list.cs#L85-L109

LibGit2Sharp/Index.cs
```diff
@@ -92,17 +92,12 @@ public virtual bool IsFullyMerged
...
+        private IEnumerable<IndexEntry> AllIndexEntries()
         {
-            var entryCount = Count;
-            var list = new List<IndexEntry>(entryCount);
-
-            for (int i = 0; i < entryCount; i++)
+            for (long i = 0; i < Count; i++)
             {
-                list.Add(this[i]);
+                yield return this[i];
```
Therzok: This is the line I was talking about. Are we _really_ sure we go over int bounds?

Therzok:
Okay, I want to expand a bit more here, since I feel guilty of not explaining enough.

If we do end up with a count higher than int, the error will end up in _usercode_ when the user transforms the IEnumerable into another type of collection.

I'm not sure we want that. I'm not sure on how to fix this thing either. We could probably make our own collection implementation to handle long sizes.

Probably make it contain two Int.MaxValue arrays as underlying holder.

nulltoken:
@Therzok Thanks for the details. Below my thoughts.

As far as I read the code, the current `List` is only used during the enumeration, then discarded (indeed, we don't want to actually cache the index content as it may be changed behind the scene by a different process). Each call to `GetEnumerator()` recreates a new `List`. In order to build this list, we enumerate the whole content of the index. The proposed change allows the user to break from the enumeration, without retrieving the whole content of the index. From a performance standpoint, I'm not sure there's a clear gain with favoring the `List`.

>If we do end up with a count higher than int, the error will end up in usercode when the user transforms the IEnumerable into another type of collection.

Wow! Nice corner-case :+1: I can't think of a proper solution to help the caller with that, would the enumeration yield more than `int.MaxValue` results. We could add some xml doc though to `Count` and `GetEnumerator()`.

jamill:
> As far as I read the code, the current List is only used during the enumeration

I agree - I don't think this change really affects how we store / access the index internally.

> we don't want to actually cache the index content as it may be changed behind the scene by a different process

Is the current access pattern actually safe against the index being modified by an external process? Wouldn't we get torn reads of the index? If the count of elements in the index changes, we would either attempt to read off the end of the index (or stop reading before we have enumerated all elements)?

> If we do end up with a count higher than int, the error will end up in usercode when the user transforms the IEnumerable into another type of collection.

using a long does make it a bit more cumbersome for consumers use if they wanted to store the index entries in an array, but it does reflect the underlying type. If there is an index that exceeds 2GB, some layer has to deal with it. libgit2 handles it, so either LibGit2Sharp would fail, or the consuming app would have to handle it. I think I prefer to have LibGit2Sharp be able to handle this possibility. It is possible to create arrays with more than 2GB elements, but it requires some extra flags (see [array] (https://msdn.microsoft.com/en-us/library/system.array%28v=vs.110%29.aspx) and [gcAllowVeryLargeObjects] (https://msdn.microsoft.com/en-us/library/hh285054(v=vs.110).aspx)), but I suspect there might be other problems working with indexes / arrays this large.

@nulltoken - In my rebase pr, I wrote a little helper [extension] (https://github.com/libgit2/libgit2sharp/pull/964/files#diff-1915aa8609ce88f027db96e5b5ca8e70R3614) for converting values to a long, if you think that is appropriate.

nulltoken:
>Is the current access pattern actually safe against the index being modified by an external process? Wouldn't we get torn reads of the index? If the count of elements in the index decreases, wouldn't we attempt to read past the end of the index?

@jamill Good point! Maybe could we get a snapshot of the index (`git_index_dup`?) and iterate over this snpashot.

>@nulltoken - In my rebase pr, I wrote a little helper extension for converting values to a long, if you think that is appropriate.

:heart: it!

ethomson:
So I think this whole thing is probably a bad idea, no?  It was pointed out to me that you can't have a .NET array that has more than `Int.MaxValue` values, at least without some magical other setting.  So this whole idea seems wrongheaded.

Should we just drop this entirely and keep what we have?

jamill:
The bit of work I would like to keep is the common code to convert these `size_t`  values down to an int. That can be a separate issue.

nulltoken:
>The bit of work I would like to keep is the common code to convert these `size_t`  values down to an int. That can be a separate issue.

Or... I can downsize the PR to just those changes...

nulltoken:
@jamill @ethomson Ready for review

nulltoken:
As we're no longer breaking the API, dropping the `stabilization` label

jamill:
This looks good to me

PR #1029, opened 2015-04-21
Fail gracefully when tree commit create fails

aroden-salesforce:
Fix 1028

### DIFF
--- a/LibGit2Sharp/TreeDefinition.cs
+++ b/LibGit2Sharp/TreeDefinition.cs
@@ -310,7 +310,12 @@ namespace LibGit2Sharp
                 builtTreeEntryDefinitions.ForEach(t => entries[t.Item1] = t.Item2);
 
                 ObjectId treeId = builder.Write();
-                return repository.Lookup<Tree>(treeId);
+                var result = repository.Lookup<Tree>(treeId);
+                if (result == null)
+                {
+                    throw new LibGit2SharpException("Unable to read created tree");
+                }
+                return result;
             }
         }
 


LibGit2Sharp/TreeDefinition.cs
```diff
@@ -310,7 +310,12 @@ internal Tree Build(Repository repository)
                 builtTreeEntryDefinitions.ForEach(t => entries[t.Item1] = t.Item2);
 
                 ObjectId treeId = builder.Write();
-                return repository.Lookup<Tree>(treeId);
+                var result = repository.Lookup<Tree>(treeId);
+                if (result == null)
+                {
+                    throw new LibGit2SharpException("Created tree not found! Can occur because antivirus or repo curroption");
```
ethomson: If `builder.Write` is returning an `ObjectId`, I would expect that it exists and this test is not necessary.  If `builder.Write` is claiming that it succeeded, but did not succeed, then we should fix _that_ problem.

Why is it lying?

aroden-salesforce: builder.Write() does succeed and returns the correct created objectId.  Lookup<Tree>(id)  fails  because an anti-virus client blocks the read to the created object file. 

It is just nicer to fail gracefully here since Lookup just returns null when the object can't be found.

aroden-salesforce:
Attached find process monitor showing the object created successfully then the read denied. Further down the AV deletes the file and shows it in its log.  Libgit2sharp null pointers during this process

block_accessed

LibGit2Sharp/TreeDefinition.cs
ethomson: Hahahaha.  Well, it's pretty shocking that an anti-virus software would be so terrible as to put you in this situation.

That said, please don't make assumptions about why a tree wasn't found in the error message.  Something more like "unable to write tree" would be more appropriate.

ethomson: But thanks for clarifying, indeed I misunderstood and thought the AV software was failing the *write* not the resultant *read*.

nulltoken:
:cool: Thanks!

ethomson:
:+1:  Thanks @aroden-salesforce !

nulltoken:

PR #1072, opened 2015-06-04
Introduce SSH functionality

Therzok:
Added SshAgentCredentials for querying ssh-agent, SshUserKeyCredentials
for authenticating with a given ssh key-pair.

Introduced UsernameQueryCredentials which returns the supported credential
types.

Authentication exceptions are now translated from libgit2.

Supersedes #852

### DIFF
--- a/LibGit2Sharp/AuthenticationException.cs
+++ b/LibGit2Sharp/AuthenticationException.cs
@@ -1,0 +1,55 @@
+﻿using System;
+using System.Runtime.Serialization;
+using LibGit2Sharp.Core;
+
+namespace LibGit2Sharp
+{
+    /// <summary>
+    /// The exception that is thrown when an operation which requires an
+    /// authentication fails.
+    /// </summary>
+    [Serializable]
+    public class AuthenticationException : LibGit2SharpException
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="LibGit2Sharp.AuthenticationException"/> class.
+        /// </summary>
+        public AuthenticationException()
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="LibGit2Sharp.AuthenticationException"/> class with a specified error message.
+        /// </summary>
+        /// <param name="message">A message that describes the error.</param>
+        public AuthenticationException(string message)
+            : base(message)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="LibGit2Sharp.AuthenticationException"/> class with a specified error message and a reference to the inner exception that is the cause of this exception.
+        /// </summary>
+        /// <param name="message">The error message that explains the reason for the exception.</param>
+        /// <param name="innerException">The exception that is the cause of the current exception. If the <paramref name="innerException"/> parameter is not a null reference, the current exception is raised in a catch block that handles the inner exception.</param>
+        public AuthenticationException(string message, Exception innerException)
+            : base(message, innerException)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="LibGit2Sharp.AuthenticationException"/> class with a serialized data.
+        /// </summary>
+        /// <param name="info">The <see cref="SerializationInfo"/> that holds the serialized object data about the exception being thrown.</param>
+        /// <param name="context">The <see cref="StreamingContext"/> that contains contextual information about the source or destination.</param>
+        protected AuthenticationException(SerializationInfo info, StreamingContext context)
+            : base(info, context)
+        {
+        }
+
+        internal AuthenticationException(string message, GitErrorCode code, GitErrorCategory category)
+            : base(message, code, category)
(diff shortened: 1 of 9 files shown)


Therzok:
Taking some important bits into discussion:

* libgit2 removed support for libssh2 embedding. What is the best way to make cmake look for libssh2 in a relative directory if it's not installed system-wide?

* I'll have to update nativebinaries scripts to be able to deploy these things once I find out the above. So we can ship libssh2 + libgit2 binaries in one package! =D

* We need some test on these things. Taken from previous PR:
> The ssh tests are a bit of a hassle to set up, but you can copy the code from libgit2's Travis script. I would recommend having at least one test run there to make sure basic functionality doesn't get broken.

* Once we have a combination of packages, we can test both without and with SSH support, by conditionally checking that things throw a `InvalidOperationException` (probably should be a `NotSupportedException`) on querying SSH credentials when native binaries don't have support for that.

Therzok:
cc @nulltoken

Therzok:
I'll rebase this as soon as possible. And I would like to have this reviewed though.

dannyzhan:
Hi Guys, how's going with ssh support in libgit2sharp?
I really need this feature in my project.
Thanks!

Therzok:
@nulltoken It's rebased atm. The only problem is getting the libssh2 native binaries alongside libgit2 ones.

carlosmn:
That part is what https://github.com/carlosmn/libgit2sharp/tree/cmn/managed-ssh was trying to bridge by not needing libssh2, but it would require our own version of the ssh library currently.

Therzok:
Going down the route of having _that_ managed ssh implementation can cause some issues such as:
 * We're still shipping a ssh library, so the whole thing is still shipping a dependency (one of your arguments).
 * libssh2 could be compiled to target mobile apps and as a result, making libgit2sharp a PCL could be easier.

This could work for now, I'm not saying no. If a PCL for SSH appears, it's easy to migrate. Just stating the available options.

carlosmn:
*We* wouldn't be the vendors for that library, it's a dependency the same way libssh2 is a dependency. The problems come when we're responsible for the updates, not when we use a library.

Unfortunately there doesn't seem to be an ssh library with active support, so the point may be moot as we'd have to maintain some version of it.

I don't get why having to compile an extra C library would be easier than using managed library for ssh if you wanna run it on a phone, IIRC you can't load native code on Windows Phone anyhow.

Therzok:
@carlosmn it all depends on the type of the project. In example, SSH.NET is a simple .NET 4.0 library (as opposed to being a PCL - Portable Class Library), which could then be used as a common library between all the projects you have (WinPhone, Android, iOS, etc). That means, this specific library won't work in a mobile context.

I was working on making LibGit2Sharp a PCL itself, so I could afterwards use it in a mobile application context, and ssh support would come via libssh2.

CumpsD:
What's the status on this? What exactly is blocking SSH support right now?

carlosmn:
There is still no good way to have ssh support in the libgit2 library that comes with the nuget packages. We've spent a lot of time and energy trying to find a solution to us becoming vendors of a crypto library, but so far no solution has presented itself. This has left little time to worry about landing this.

CumpsD:
Sadly enough, I went with NGit to get SSH working, would've loved to use the cleaner API of libgit2sharp

EvAlex:
Any updates on this? Crucial feature. The whole library is of no use without it.

carlosmn:
SSH support is in the bindings; the shipped libgit2 doesn't because we do not want to ship third-party crypto software in our packages.

If there are missing features from this PR, please open a new one adding that.

Therzok:
@carlosmn We should at least surface the API using feature checks.

carlosmn:
libgit2 knows whether it was built with ssh support so it'll error out on its own. I'm not a fan of the idea of the bindings doing feature checks themselves.

EvAlex:
Should there at least be a paragraph with all this info in README.md? I assure you, this matter is not absolutely clear to everyone willing to use this library. It would assist people a lot, if there was some guidance on ssh hit checkout.

carlosmn:
Without a description (and a reasonable way) of how to use your own library, I don't know how useful that'd be.

ssougnez:
I'm confused... Is it supported or not ? I'm trying to clone a repository using SSH but I really can't find the way to do so. Some comments on the project tends to say it is supported, other say the opposite. Is it possible to have a clear statement on this please? 

I really think that issue should be addressed ASAP... Who wants to use password in their code while a SSH key can be used? Without this feature, this awesome library is quite limited.

Thanks a lot for your work

leobuskin:
@ssougnez you can use libgit2sharp fork with SSH support based on @Therzok work, available in NuGet under 'LibGit2Sharp-SSH' name.

rebizu:
@leobuskin what version is it a fork of ?

dannyzhan:
sadly, I have to use git command directly to support SSH.

PR #1123, opened 2015-06-26
Allow libgit2# to redirect the global/system/xdg config search path

rcorre:
This provides an interface for libgit2# to redirect config file lookups via libgit2's `git_libgit2_opt`.

This is useful for unit-testing methods that hit the global, system, or xdg config files.

It is currently possible to redirect config access using `RepositoryOptions`, but this only works if the functions you are testing take a `Repository` as an argument. This isn't much of a problem for libgit2#, but other frameworks may not want to pass a `Repository` all the way through their call stack to support testing with a faked config. This provides a more foolproof way to sandbox your tests as it guarantees that all config access is redirected (libgit2 [uses this for testing](https://github.com/libgit2/libgit2/blob/c7f94123569e8fe00ffb3a35e6a12b6ebe9320ec/tests/config/global.c#L5)).

The first commit adds the actual functionality. `git_libgit2_opts` has a number of other options -- I only added bindings for search path redirection, but this sets the stage for binding other libgit2 settings if they are needed.

The second commit shows how this could be used within libgit2# tests. The current redirection scheme can be flaky -- I noticed some tests failing because they [only redirected system config access](https://github.com/libgit2/libgit2sharp/blob/vNext/LibGit2Sharp.Tests/DiffTreeToTreeFixture.cs#L1049), and my personal global config was interfering.
The test could be fixed without this, of course, but I think this will help prevent mistakes in the long run.

[EDIT]
I've removed the second commit for now -- my partial fixup of the test fixtures was causing problems in as many tests as it fixed.

This PR just allows libgit2# consumers to redirect config access, I can work on a second PR to use this in the lg2# test fixtures if it seems like a good idea (started [here](https://github.com/rcorre/libgit2sharp/commit/f5002458f9e8a8867a86c89e5c297678c084d650)).

### DIFF
--- a/LibGit2Sharp.Tests/ConfigurationFixture.cs
+++ b/LibGit2Sharp.Tests/ConfigurationFixture.cs
@@ -431,5 +431,127 @@ namespace LibGit2Sharp.Tests
                 Assert.Null(signature);
             }
         }
+
+        [Fact]
+        public void CanSetAndGetSearchPath()
+        {
+            string globalPath = Path.Combine(Constants.TemporaryReposPath, Path.GetRandomFileName());
+            string systemPath = Path.Combine(Constants.TemporaryReposPath, Path.GetRandomFileName());
+            string xdgPath    = Path.Combine(Constants.TemporaryReposPath, Path.GetRandomFileName());
+
+            GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Global, globalPath);
+            GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.System, systemPath);
+            GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Xdg,    xdgPath);
+
+            Assert.Equal(globalPath, GlobalSettings.GetConfigSearchPaths(ConfigurationLevel.Global).Single());
+            Assert.Equal(systemPath, GlobalSettings.GetConfigSearchPaths(ConfigurationLevel.System).Single());
+            Assert.Equal(xdgPath,    GlobalSettings.GetConfigSearchPaths(ConfigurationLevel.Xdg).Single());
+
+            // reset the search paths to their defaults
+            GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Global, null);
+            GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.System, null);
+            GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Xdg,    null);
+        }
+
+        [Fact]
+        public void CanSetAndGetMultipleSearchPaths()
+        {
+            string[] paths =
+            {
+                Path.Combine(Constants.TemporaryReposPath, Path.GetRandomFileName()),
+                Path.Combine(Constants.TemporaryReposPath, Path.GetRandomFileName()),
+                Path.Combine(Constants.TemporaryReposPath, Path.GetRandomFileName()),
+            };
+
+            GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Global, paths);
+
+            Assert.Equal(paths, GlobalSettings.GetConfigSearchPaths(ConfigurationLevel.Global));
+
+            // set back to the defaults
+            GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Global, null);
+        }
+
+        [Fact]
+        public void CanResetSearchPaths()
+        {
+            // all of these calls should reset the config path to the default
+            Action[] resetActions =
+            {
+                () => GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Global),
+                () => GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Global, null),
+                () => GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Global, string.Empty),
+                () => GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Global, new string[] { }),
+            };
+
+            // record the default search path
+            GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Global, null);
+            var oldPaths = GlobalSettings.GetConfigSearchPaths(ConfigurationLevel.Global);
+            Assert.NotNull(oldPaths);
+
+            // generate a non-default path to set
+            var newPaths = new string[] { Path.Combine(Constants.TemporaryReposPath, Path.GetRandomFileName()) };
+
+            foreach (var tryToReset in resetActions)
+            {
+                // change to the non-default path
+                GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Global, newPaths);
+                Assert.Equal(newPaths, GlobalSettings.GetConfigSearchPaths(ConfigurationLevel.Global));
+
+                // set it back to the default
+                tryToReset();
+                Assert.Equal(oldPaths, GlobalSettings.GetConfigSearchPaths(ConfigurationLevel.Global));
+            }
+
+            // make sure the config paths are reset after the test ends
+            GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Global, null);
+        }
+
+        [Fact]
+        public void CanAppendToSearchPaths()
+        {
+            string appendMe = Path.Combine(Constants.TemporaryReposPath, Path.GetRandomFileName());
+            var prevPaths = GlobalSettings.GetConfigSearchPaths(ConfigurationLevel.Global);
+
+            // append using the special name $PATH
+            GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Global, "$PATH", appendMe);
+
+            var currentPaths = GlobalSettings.GetConfigSearchPaths(ConfigurationLevel.Global);
+            Assert.Equal(currentPaths, prevPaths.Concat(new[] { appendMe }));
+
+            // set it back to the default
+            GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Global, null);
+        }
+
+        [Fact]
+        public void CanRedirectConfigAccess()
+        {
+            var scd1 = BuildSelfCleaningDirectory();
+            var scd2 = BuildSelfCleaningDirectory();
+
+            Touch(scd1.RootedDirectoryPath, ".gitconfig");
+            Touch(scd2.RootedDirectoryPath, ".gitconfig");
+
+            // redirect global access to the first path
+            GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Global, scd1.RootedDirectoryPath);
+
+            // set a value in the first config
+            using (var config = Configuration.BuildFrom(null))
+            {
+                config.Set("luggage.code", 9876, ConfigurationLevel.Global);
+                Assert.Equal(9876, config.Get<int>("luggage.code", ConfigurationLevel.Global).Value);
+            }
+
+            // redirect global config access to path2
+            GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Global, scd2.RootedDirectoryPath);
+
+            // if the redirect succeeds, the value set in the prior config should not be visible
+            using (var config = Configuration.BuildFrom(null))
+            {
+                Assert.Equal(-1, config.GetValueOrDefault<int>("luggage.code", ConfigurationLevel.Global, -1));
+            }
+
+            // reset the search path to the default
+            GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Global, null);
+        }
     }
 }
--- a/LibGit2Sharp/Core/NativeMethods.cs
+++ b/LibGit2Sharp/Core/NativeMethods.cs
@@ -581,6 +581,25 @@ namespace LibGit2Sharp.Core
         [DllImport(libgit2)]
         internal static extern int git_libgit2_features();
 
+        #region git_libgit2_opts
+
+        // Bindings for git_libgit2_opts(int option, ...):
+        // Currently only GIT_OPT_GET_SEARCH_PATH and GIT_OPT_SET_SEARCH_PATH are supported,
+        // but other overloads could be added using a similar pattern.
+        // CallingConvention.Cdecl is used to allow binding the the C varargs signature, and each possible call signature must be enumerated.
+        // __argslist was an option, but is an undocumented feature that should likely not be used here.
+
+        // git_libgit2_opts(GIT_OPT_GET_SEARCH_PATH, int level, git_buf *buf)
+        [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern int git_libgit2_opts(int option, uint level, GitBuf buf);
+
+        // git_libgit2_opts(GIT_OPT_SET_SEARCH_PATH, int level, const char *path)
+        [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern int git_libgit2_opts(int option, uint level,
+            [MarshalAs(UnmanagedType.CustomMarshaler, MarshalCookie = UniqueId.UniqueIdentifier, MarshalTypeRef = typeof(StrictUtf8Marshaler))]string path);
+
+        #endregion
+
         [DllImport(libgit2)]
         internal static extern int git_graph_ahead_behind(out UIntPtr ahead, out UIntPtr behind, RepositorySafeHandle repo, ref GitOid one, ref GitOid two);
 
--- a/LibGit2Sharp/Core/Proxy.cs
+++ b/LibGit2Sharp/Core/Proxy.cs
@@ -3193,6 +3193,60 @@ namespace LibGit2Sharp.Core
             return (BuiltInFeatures)NativeMethods.git_libgit2_features();
         }
 
+        // C# equivalent of libgit2's git_libgit2_opt_t
+        private enum LibGitOption
+        {
+            GetMWindowSize,        // GIT_OPT_GET_MWINDOW_SIZE
+            SetMWindowSize,        // GIT_OPT_SET_MWINDOW_SIZE
+            GetMWindowMappedLimit, // GIT_OPT_GET_MWINDOW_MAPPED_LIMIT
+            SetMWindowMappedLimit, // GIT_OPT_SET_MWINDOW_MAPPED_LIMIT
+            GetSearchPath,         // GIT_OPT_GET_SEARCH_PATH
+            SetSearchPath,         // GIT_OPT_SET_SEARCH_PATH
+            SetCacheObjectLimit,   // GIT_OPT_SET_CACHE_OBJECT_LIMIT
+            SetCacheMaxSize,       // GIT_OPT_SET_CACHE_MAX_SIZE
+            EnableCaching,         // GIT_OPT_ENABLE_CACHING
+            GetCachedMemory,       // GIT_OPT_GET_CACHED_MEMORY
+            GetTemplatePath,       // GIT_OPT_GET_TEMPLATE_PATH
+            SetTemplatePath,       // GIT_OPT_SET_TEMPLATE_PATH
+            SetSslCertLocations,   // GIT_OPT_SET_SSL_CERT_LOCATIONS
+        }
+
+        /// <summary>
+        /// Get the paths under which libgit2 searches for the configuration file of a given level.
+        /// </summary>
+        /// <param name="level">The level (global/system/XDG) of the config.</param>
+        /// <returns>
+        ///     The paths delimited by 'GIT_PATH_LIST_SEPARATOR' (<see cref="GlobalSettings.PathListSeparator"/>).
+        /// </returns>
+        public static string git_libgit2_opts_get_search_path(ConfigurationLevel level)
+        {
+            string path;
+
+            using (var buf = new GitBuf())
+            {
+                var res = NativeMethods.git_libgit2_opts((int)LibGitOption.GetSearchPath, (uint)level, buf);
+                Ensure.ZeroResult(res);
+
+                path = LaxUtf8Marshaler.FromNative(buf.ptr) ?? string.Empty;
+            }
+
+            return path;
+        }
+
+        /// <summary>
+        /// Set the path(s) under which libgit2 searches for the configuration file of a given level.
+        /// </summary>
+        /// <param name="level">The level (global/system/XDG) of the config.</param>
+        /// <param name="path">
+        ///     A string of paths delimited by 'GIT_PATH_LIST_SEPARATOR' (<see cref="GlobalSettings.PathListSeparator"/>).
+        ///     Pass null to reset the search path to the default.
+        /// </param>
+        public static void git_libgit2_opts_set_search_path(ConfigurationLevel level, string path)
+        {
+            var res = NativeMethods.git_libgit2_opts((int)LibGitOption.SetSearchPath, (uint)level, path);
+            Ensure.ZeroResult(res);
+        }
+
         #endregion
 
         private static ICollection<TResult> git_foreach<T, TResult>(
--- a/LibGit2Sharp/GlobalSettings.cs
+++ b/LibGit2Sharp/GlobalSettings.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Reflection;
+using System.Collections.Generic;
 using LibGit2Sharp.Core;
 
 namespace LibGit2Sharp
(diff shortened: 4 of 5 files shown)


LibGit2Sharp/Core/NativeMethods.cs
```diff
@@ -581,6 +581,32 @@ static NativeMethods()
...
+        // git_libgit2_opts(GIT_OPT_GET_SEARCH_PATH, int level, git_buf *buf)
+        [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern int git_libgit2_opts(int option, uint level, GitBuf buf);
+
+        // git_libgit2_opts(GIT_OPT_GET_SEARCH_PATH, int level, git_buf *buf)
+        [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern int git_libgit2_opts(int option, uint level,
+            [MarshalAs(UnmanagedType.CustomMarshaler, MarshalCookie = UniqueId.UniqueIdentifier, MarshalTypeRef = typeof(StrictUtf8Marshaler))]string name);
+
+        // Not yet implemented libgit2_opts signatures:
```
ethomson: We could definitely change the functionality in libgit2 and this comment would get out of date.  This might be frustrating to a future implementer, so I would recommend dropping these comments.  Maybe we should have something alluding to the fact that there are other possible signatures that the above do not cover.

LibGit2Sharp/Core/NativeMethods.cs
```diff
@@ -581,6 +581,32 @@ static NativeMethods()
...
 
+        // Bindings for git_libgit2_opts(int option, ...):
+        // Varargs can be bound using __arglist in .NET, but I had trouble getting that working with Mono.
+        // Instead of using __arglist, I enumerate the possible signatures here:
+
+        // git_libgit2_opts(GIT_OPT_GET_SEARCH_PATH, int level, git_buf *buf)
+        [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern int git_libgit2_opts(int option, uint level, GitBuf buf);
+
+        // git_libgit2_opts(GIT_OPT_GET_SEARCH_PATH, int level, git_buf *buf)
```
Therzok: Shouldn't this be `GIT_OPT_SET_SEARCH_PATH, int level, const char *path` ?

rcorre: yep, copy-paste error strikes again

LibGit2Sharp/Core/NativeMethods.cs
rcorre: Good point, I guess someone wanting to implement another one of these would have checked the libgit2 docs anyways

nulltoken:
Your proposal is growing on me.

Actually, I'm starting to wonder if we shouldn't obsolete
`RepositoryOptions.{[Global|Xdg|System]ConfigurationLocation}` in favor of this.

Thoughts?

/cc @shitkey @ammeep @dahlbyk

rcorre:
I have a branch [here](https://github.com/rcorre/libgit2sharp/tree/test-fixes) where I use this for the libgit2# tests. With that in place, I should be able to strip out all uses of `ConfigurationLocation` in the tests.

As a bonus, it prevents tests from accidentally writing to a user's config or taking a dependency on the test environment (e.g. failing if you don't have the right `core.autocrlf` setting).

nulltoken:
@rcorre :cool:

We would also have to cleanup all the tests that actually rely on those RepositoryOptions properties as well. However, before diving into this, I'd like to make sure that it's ok to set this globally and that there's no need for a temporary override of those ones.

whoisj:
> Actually, I'm starting to wonder if we shouldn't obsolete
RepositoryOptions.{[Global|Xdg|System]ConfigurationLocation} in favor of this.

Given the desire to be able to, eventually, run unit-tests in parallel this would be a bummer. It would be nice if the `RepositoryOptions` overrode the values still.

nulltoken:
> Given the desire to be able to, eventually, run unit-tests in parallel this would be a bummer. It would be nice if the `RepositoryOptions` overrode the values still.

I've pondered this as well, but I think we're already doomed (filters, subtransports, ...). Besides, if the only use of those `RepositoryOptions` properties is to help us out with the tests, that may be interesting hint that we should deprecate them. :wink:

whoisj:
> I've pondered this as well, but I think we're already doomed (filters, subtransports, ...). Besides, if the only use of those RepositoryOptions properties is to help us out with the tests, that may be interesting hint that we should deprecate them. :wink:

touché

nulltoken:
@shiftkey Unkess I'm wrong, you're making use of this feature. How do you feel about @rcorre's proposal?

shiftkey:
@nulltoken I'm interested in this proposal.

LibGit2Sharp.Tests/ConfigurationFixture.cs
```diff
@@ -409,5 +409,94 @@ public void PassingANonExistingLocalConfigurationFileToBuildFromthrowss()
...
+            Touch(scd1.RootedDirectoryPath, ".gitconfig");
+            Touch(scd2.RootedDirectoryPath, ".gitconfig");
+
+            // redirect global access to the first path
+            GlobalSettings.SetConfigSearchPath(ConfigurationLevel.Global, scd1.RootedDirectoryPath);
+
+            // set a value in the first config
+            using (var config = Configuration.BuildFrom(null))
+            {
+                config.Set("luggage.code", 9876, ConfigurationLevel.Global);
```
nulltoken: ZOMG! You're disclosing secrets, here!!1! :stuck_out_tongue_winking_eye:

LibGit2Sharp/GlobalSettings.cs
```diff
@@ -203,5 +204,48 @@ public static void DeregisterFilter(FilterRegistration registration)
             Proxy.git_filter_unregister(registration.Name);
             registration.Free();
         }
+
+        /// <summary>
+        /// The separator used in path list strings (like in the PATH environment variable).
+        /// A semi-colon ";" is used on Windows, and a colon ":" for all other systems.
+        /// </summary>
+        public static char PathListSeparator
```
nulltoken: Do we need to expose this?

rcorre: Probably not, I can go ahead and make that private

Therzok: https://msdn.microsoft.com/en-us/library/system.io.path.pathseparator(v=vs.110).aspx ?

Therzok: It's also returning the right thing for  [https://github.com/mono/mono/blob/88d2b9da2a87b4e5c82abaea4e5110188d49601d/mono/metadata/file-io.c#L1132-L1140](Mono).

rcorre:
Removed `GlobalSettings.PathListSeparator` in favor of `Path.PathListSeparator`. Thanks @Therzok!

LibGit2Sharp.Tests/ConfigurationFixture.cs
whoisj: Isn't that supposed to be "1-2-3-4-5?"?

I mean "That's the stupidest combination I've ever heard of in my life! That's the kinda thing an idiot would have on his luggage!"

:stuck_out_tongue:

shiftkey: Shout out to obscure Mel Brooks references...

carlosmn: Literally that first thing that came to mind when I read that line.

dahlbyk:
What's the expected/actual behavior for `GlobalSettings.SetConfigSearchPaths(configLevel, Enumerable.Empty<string>())`?

rcorre:
It _should_ restore the default search path. Having trouble getting libgit2# to build right now, but I'll verify that when I can.

rcorre:
Confirmed that `""` and `Enumerable.Empty<string>()` behave as expected (same as `null`).
As an aside, would overloading `SetConfigSearchPath` be preferable to having `SetConfigSearchPath` and `SetConfigSearchPaths`?

dahlbyk:
> Confirmed that `""` and `Enumerable.Empty<string>()` behave as expected (same as `null`).

Add tests to prove the behavior?

> As an aside, would overloading `SetConfigSearchPath` be preferable to having `SetConfigSearchPath` and `SetConfigSearchPaths`?

I'd be fine with an overloaded `SetConfigSearchPath`.

nulltoken:
>I'd be fine with an overloaded SetConfigSearchPath

Maybe in the plural form?

rcorre:
`SetCofigSearchPaths` is technically what it does, though I'd imagine setting a single path is the most common case.

dahlbyk:
`SetCofigSearchPaths` it is, then

Therzok:
`Cofigs` are really nice things to have.

rcorre:
Overloading for `IEnumerable`/`string` wouldn't work as `null` would be an ambiguous argument, so I compromised with:
```
SetConfigSearchPaths(ConfigurationLevel level, params string[] paths)
```
I noticed that libgit2 documents that the special string "$PATH" refers to the current path, so I documented an tested that too (as well as the empty string behavior).

As tempting as it was to shave another character off of the function name, I decided to keep the `n` in `Config` :p

dahlbyk:
> Overloading for `IEnumerable`/`string` wouldn't work as `null` would be an ambiguous argument, so I compromised...

:+1:

> I noticed that libgit2 documents that the special string "$PATH" refers to the current path, so I documented and tested that too (as well as the empty string behavior).

:+1: 

> As tempting as it was to shave another character off of the function name, I decided to keep the `n` in `Config` :p

:+1: 

> The Travis CI build failed

:-1:

ethomson:
> The Travis CI build failed

I kicked it, but Travis is again having Mac Issues.

rcorre:
Thanks!

Incidentally, should LibGit2Sharp build with `mono >= 4.0`?
I tried to test the linux build locally, but I had to roll back mono to `3.12`

dahlbyk:
> Incidentally, should LibGit2Sharp build with mono >= 4.0?

@nulltoken @Therzok?

Therzok:
This should make it build? https://github.com/libgit2/libgit2sharp/pull/1034

rcorre:
> This should make it build? #1034

Ah, I wasn't reading the output closely enough. It actually finished building, but xunit was failing with some `PrivateInvoke` call as soon as it tried to run the test suite.
Either way, its not really related to this PR, but I'll open a separate issue if it looks like LibGit2#'s fault.

dahlbyk:
Looks like we're good here? Will merge my next time through unless @nulltoken beats me to it.

nulltoken:
>Actually, I'm starting to wonder if we shouldn't obsolete
`RepositoryOptions.{[Global|Xdg|System]ConfigurationLocation}` in favor of this.

@dahlbyk Given this PR, I can't think of a use case where we would need a way to override this at the repo level. What's your take about this?

> I noticed that libgit2 documents that the special string "$PATH" refers to the current path, so I documented and tested that too (as well as the empty string behavior).

I'm not sure about this. `$PATH` reads like an environment variable and may be confusing. How about accepting relative paths and transforming them into full paths before passing them to libgit2?

rcorre:
> I'm not sure about this. $PATH reads like an environment variable and may be confusing (besides being not very .NET friendly/standard)

This is how libgit2 handles it, so I don't have much choice here. If someone passes "$PATH", it will be expanded to the current search path, so I figured its less surprising if it is at least documented on the libgit2# end.

> How about accepting relative paths and transforming them into full paths before passing them to libgit2?

$PATH has nothing to do with relative paths -- it expands to the current config search path (e.g. ~/.gitconfig).
In other words, the following are equivalent:
```SetConfigSearchPaths("$PATH", "foo")```
```SetConfigSearchPaths(GetConfigSearchPaths, "foo")```

>> Actually, I'm starting to wonder if we shouldn't obsolete
RepositoryOptions.{[Global|Xdg|System]ConfigurationLocation} in favor of this.

>@dahlbyk Given this PR, I can't think of a use case where we would need a way to override this at the repo level. What's your take about this?

I think @whoisj made a good point about multithreaded tests. While they may not be in the books for lg2#, consumers may want to leverage per-repo config redirection.

This PR implements a more coarse-grained solution, but the advantage is you can use it even if you are reading a global config without a repo instance.

**Comment by dahlbyk** [comment]
> > I'm not sure about this. `$PATH` reads like an environment variable and may be confusing (besides being not very .NET friendly/standard)
>
> This is how libgit2 handles it, so I don't have much choice here. If someone passes "$PATH", it will be expanded to the current search path, so I figured its less surprising if it is at least documented on the libgit2# end.

:+1: even if unlikely to be stumbled upon accidentally. Let's call it a "free" microoptimization to avoid marshaling and splitting the current list of paths?

> > Actually, I'm starting to wonder if we shouldn't obsolete
> > `RepositoryOptions.{[Global|Xdg|System]ConfigurationLocation}` in favor of this.
>
> @dahlbyk Given this PR, I can't think of a use case where we would need a way to override this at the repo level. What's your take about this?
> I think @whoisj made a good point about multithreaded tests. While they may not be in the books for lg2#, consumers may want to leverage per-repo config redirection.
> 
> This PR implements a more coarse-grained solution, but the advantage is you can use it even if you are reading a global config without a repo instance.

The more compelling case to me is the plug-in model, e.g. the GitHub VS extension may want to load in config for its credential store without impacting anything else that depends on the bundled LG2S. Maybe not a common case, but I don't know that it hurts us to support both. Maybe update docs for each to `<seealso .../>`?

**Comment by rcorre** [comment]
I've added `<seealso>` to cross-reference `SetConfigSearchPaths` and `RepsitoryOptions`.

**Comment by nulltoken** [comment]
@rcorre Forget about my remark about `$PATH`, I read the code backwards. Sorry for the noise.

@shiftkey @jamill @ethomson @carlosmn Do you think we would indeed need both an app-domain level and a per repository override?

**Comment by shiftkey** [comment]
Apologies if I mess up something here, I've not kept up with this PR as it's evolved over time...

> Do you think we would indeed need both an app-domain level and a per repository override?

I guess I'm not clear on the scenario where a "per repository" override would be needed. I'm trying to think of a scenario in the past where I've come across this, but coming up empty.

:thumbsup: with the direction this is heading in - will definitely help with testing on my end...

**Comment by rcorre** [comment]
Another thing I've been looking into is libgit2's `config_backend`. If custom config backends were supported in libgit2#, could that replace the `RepositoryOptions` config path override?

Setting it up might be a bit more involved than using `RepositoryOptions`, but it could be much more flexible (i.e. faking configs in memory).

**Comment by jamill** [comment]
> The more compelling case to me is the plug-in model..

I think this is the interesting use case - a setup where multiple components depend on the same LG2S, and the individual components do not want to affect each other.

**Comment by ethomson** [comment]
> The more compelling case to me is the plug-in model, e.g. the GitHub VS extension may want to load in config for its credential store without impacting anything else that depends on the bundled LG2S.

Nobody should be using the Visual Studio bundled in Visual Studio except people internal to Visual Studio.  (I'm including the GitHub VS extension in this, since they're shipping in our box but note that as soon as 2015 was RTM'd, they are now using their own LibGit2Sharp.)

I realize that you may be using this as an example of a plug-in model type system.  I don't think that LibGit2Sharp can predict this sort of behavior and that if Visual Studio (to continue this example) were to offer consumers the ability to work with *its* LibGit2Sharp instance then there needs to be a contract between consumers and the provider about what can be called.  I don't think LibGit2Sharp can really be good at predicting what this/these things should look like.

**Comment by ethomson** [comment]
Anyway.  I'm :+1: on this change *and* on pluggable configuration backends.  That's something various people have wanted for a while but it has not bubbled up very high on my priority list.

File: LibGit2Sharp.Tests/ConfigurationFixture.cs [hunk 5]
```diff
@@ -409,5 +409,127 @@ public void PassingANonExistingLocalConfigurationFileToBuildFromthrowss()
...
+            GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Global, globalPath);
+            GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.System, systemPath);
+            GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Xdg,    xdgPath);
+
+            Assert.Equal(globalPath, GlobalSettings.GetConfigSearchPaths(ConfigurationLevel.Global).Single());
+            Assert.Equal(systemPath, GlobalSettings.GetConfigSearchPaths(ConfigurationLevel.System).Single());
+            Assert.Equal(xdgPath,    GlobalSettings.GetConfigSearchPaths(ConfigurationLevel.Xdg).Single());
+
+            // reset the search paths to their defaults
+            GlobalSettings.SetConfigSearchPaths(ConfigurationLevel.Global, null);
```
jamill: Do we want to try and always run the code that reset's the search paths to their defaults (via `try / finally`), to try and prevent possibly polluting other tests? Although, if there is a bug in logic for setting the global search paths, then maybe the rest logic would also not work anways...

LibGit2Sharp/Core/NativeMethods.cs
```diff
@@ -581,6 +581,25 @@ static NativeMethods()
...
+        #region git_libgit2_opts
+
+        // Bindings for git_libgit2_opts(int option, ...):
+        // Varargs can be bound using __arglist in .NET, but I had trouble getting that working with Mono.
+        // Instead of using __arglist, I enumerate the possible signatures here.
+        // Currently only GIT_OPT_GET_SEARCH_PATH and GIT_OPT_SET_SEARCH_PATH are supported,
+        // but other overloads could be added using a similar pattern.
+
+        // git_libgit2_opts(GIT_OPT_GET_SEARCH_PATH, int level, git_buf *buf)
+        [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
```
jamill: Do you think a note on why you are using cdecl calling convention here would be useful (or being more explicit in the above comment)? (because of the varargs in the native library?)

LibGit2Sharp/Core/NativeMethods.cs
```diff
@@ -581,6 +581,25 @@ static NativeMethods()
...
+        internal static extern int git_libgit2_opts(int option, uint level, GitBuf buf);
+
+        // git_libgit2_opts(GIT_OPT_SET_SEARCH_PATH, int level, const char *path)
+        [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
```
ethomson: Why `CallingConvention.Cdecl`?

LibGit2Sharp/Core/Proxy.cs
```diff
@@ -3148,6 +3148,60 @@ public static BuiltInFeatures git_libgit2_features()
             return (BuiltInFeatures)NativeMethods.git_libgit2_features();
         }
 
+        // C# equivalent of libgit2's git_libgit2_opt_t
+        private enum LibGitOption
+        {
+            GetMWindowSize,        // GIT_OPT_GET_MWINDOW_SIZE
```
jamill: Did you want to explicitly set the values on the enum to match what is in lg2, just to remove any possibly ambiguity? I think this is how most of our other enums are defined.

LibGit2Sharp/Core/NativeMethods.cs
rcorre: as @jamill mentioned, this was (seemingly) necessary to bind to a varargs function. I'll include a note in the comments about that.

LibGit2Sharp/GlobalSettings.cs
```diff
@@ -203,5 +204,32 @@ public static void DeregisterFilter(FilterRegistration registration)
             Proxy.git_filter_unregister(registration.Name);
             registration.Free();
         }
+
+        /// <summary>
+        /// Get the paths under which libgit2 searches for the configuration file of a given level.
+        /// </summary>
+        /// <param name="level">The level (global/system/XDG) of the config.</param>
+        /// <returns>The paths that are searched</returns>
+        public static IEnumerable<string> GetConfigSearchPaths(ConfigurationLevel level)
```
jamill: Do you think returning the array here might be more useful for consumers? We are returning a copy of the data, so I don't think we need the "read only" property of the IEnumerable... just a thought...

LibGit2Sharp.Tests/ConfigurationFixture.cs
rcorre: Yeah, probably a good idea. As you point out, this test is a bit weird -- if the logic is flawed, the whole setup/teardown process is bugged.

LibGit2Sharp/Core/NativeMethods.cs
ethomson: Thanks!  This is (I think) the first time we've changed the decoration like this, so I think it would be worthwhile to document why.

jamill:
Thanks @rcorre - I think this looks good!

LibGit2Sharp/GlobalSettings.cs
rcorre: Yeah, forgot that `Split` returns an `[]` already, so I might as well be explicit about it.Good catch!

nulltoken: I'm not sure we return that many arrays API wide. Would there be any trouble sticking with an IEnumerable?

LibGit2Sharp/Core/Proxy.cs
rcorre: lg2 doesn't explicitly set any values, it just uses the default enum numbering, so I figured it was best to just match lg2 (I'm guessing its highly unlikely that either language is going to change the default numbering anytime soon).

jamill: sounds good

LibGit2Sharp/GlobalSettings.cs
rcorre: On one hand, array is more flexible: I can assign a `string[]` to an `IEnumerable<string>` but not visa-versa. This makes it clear on what you are getting.
On the other, `IEnumerable<string>` makes it easier to avoid breaking API changes by providing a more loose contract.
In that light, I guess I'd prefer sticking with `IEnumerable`.

jamill: Sounds good!

LibGit2Sharp.Tests/ConfigurationFixture.cs
dahlbyk: I'm generally opposed to asserts in setup/teardown, but this might be a case where we want to capture global config state at the fixture level, and then restore (and assert restoration) on teardown. If any of that fails, we're probably not going to trust these config tests anyway.

LibGit2Sharp/GlobalSettings.cs
dahlbyk: :+1: for immutable `IEnumerable`

LibGit2Sharp.Tests/ConfigurationFixture.cs
rcorre: Right now its only being used in the tests I added, the other `ConfigurationFixture` tests, still rely on the old redirection method (`RepositoryOptions` set up through a `BuildFakeConfigs` test method).

I _could_ use this redirection for all the config tests (and potentially all tests in the library), but I think that's a whole different PR :)

rcorre:
This fell pretty far behind, so I've rebased back up to vNext (and added a comment about `Cdecl`).

LibGit2Sharp/GlobalSettings.cs
Therzok: `IEnumerable`, you have my sword.

rcorre:
Anything preventing this from going through at this point?

As far as whether to deprecate `RepositoryOptions.{[Global|Xdg|System]ConfigurationLocation}`:

1. That decision could be made separately from this PR
2. I am partway through implementing `config_backend` support, which could truly replace it.

nulltoken:
@rcorre Let's do this. Could you please squash this PR into a single commit so we can merge it?

rcorre:
@nulltoken: squashed and ready to go

Therzok:
I am sad Cofigs are no longer an option. :(

nulltoken:
> @nulltoken: squashed and ready to go

@rcorre :heart: Let's merge it. Awesome job!

> I am sad Configs are no longer an option. :(

@Therzok I'm going to open an issue discussing the potential deprecation of those

Therzok:
@nulltoken you do realise I actually meant to type `Cofigs`, right?

https://github.com/libgit2/libgit2sharp/pull/1123#issuecomment-125179402

nulltoken:
@Therzok :fearful: AAAaaaaaahhhhh! ... Nope. I didn't. Thanks for being on the lookout :heart: 

@rcorre How would you feel about a tiny PR to fix this up? :pray:

nulltoken:
>@rcorre How would you feel about a tiny PR to fix this up? :pray: 

:trollface: of course :wink:

rcorre:
It _is_ one letter shorter...

nulltoken:
>It is one letter shorter...
Think of all the collective developer time we could save!

Dammit! You're right.

There are indeed a finite number of keystrokes left in developers'hands before they die.

image

Help us @shanselman you're our only hope!

PR #1131, opened 2015-07-01
Filter fixes

ethomson:
Some fixes around custom filters.

Primarily, this provides a new Create() method that filters can override, which will indicate to them when a new file is being filtered.

In our test, this allows us to delete any file that may have existed.  The problem here is that with the new racy-git handling, we re-run the filter.  This caused us to continue to append the workdir data, as if we had never seen the EOF.

Other filters, though, will likely want to know when they start filtering a file.

### DIFF
--- a/LibGit2Sharp/Filter.cs
+++ b/LibGit2Sharp/Filter.cs
@@ -113,6 +113,16 @@ namespace LibGit2Sharp
         { }
 
         /// <summary>
+        /// Indicates that a filter is going to be applied for the given file for
+        /// the given mode.
+        /// </summary>
+        /// <param name="path">The path of the file being filtered</param>
+        /// <param name="root">The path of the working directory for the owning repository</param>
+        /// <param name="mode">The filter mode</param>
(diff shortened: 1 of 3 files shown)


whoisj:
Interesting inter test dependency issues.

LibGit2Sharp/Filter.cs
```diff
@@ -113,6 +113,16 @@ protected virtual void Initialize()
         { }
 
         /// <summary>
+        /// Indicates that a filter is going to be applied for the given file for
+        /// the given mode.
+        /// </summary>
+        /// <param name="path">The path of the file being filtered</param>
+        /// <param name="root">The path of the working directory for the owning repository</param>
+        /// <param name="mode">The filter mode</param>
+        protected virtual void Create(string path, string root, FilterMode mode)
```
whoisj: This brings up a great point: if we only have a single entry-point for `Create` which takes `FilterMode` as an argument - why do we have separate `Clean` and `Smudge` entry-points?

PR #1136, opened 2015-07-02
Fixes `RebaseFixture` tests by adding .gitattributes file to repo

### LINKED ISSUE #1111: RebaseFixture Tests Assume "core.autocrlf true" but Do Not Enforce It
The `CanRebaseHandlePatchAlreadyApplied` and `VerifyRebaseDetailed` tests in the `RebaseFixture` assume that `git config core.crlf true` has been set on Windows. When this isn't true the test fail. The tests should either not assume the value or set the value.

whoisj:
Several of the tests in the `RebaseFixture` assume that `core.autocrlf = true` was set. On systems where this wasn't set to true, the tests failed. The correct solution is to not rely on `core.autocrlf` and include a `.gitattributes` file in the test repo.

Fixes #1111

### DIFF
--- a/LibGit2Sharp.Tests/FilterFixture.cs
+++ b/LibGit2Sharp.Tests/FilterFixture.cs
@@ -304,7 +304,6 @@ namespace LibGit2Sharp.Tests
             return true;
         }
 
-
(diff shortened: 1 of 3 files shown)


jamill:
:heart: Awesome! thanks!

PR #1140, opened 2015-07-07
Filter API Refactoring

whoisj:
Changed `Filter` to a short lived, per-file entity.

Standardized the exposed virtual entry-points.

Combined `Clean` and `Smudge` entry-points.

Moved all long-term constructs into `FilterRegister`.

Removed the superfluous `FilterAttributeEntry`.

Fixed up tests to use new model.


whoisj:
Mostly just a what-if, but I personally like this.

/CC @ethomson @jamill @nulltoken @ammeep

LibGit2Sharp/FilterRegistration.cs
```diff
@@ -7,44 +11,151 @@ namespace LibGit2Sharp
...
+    public abstract class FilterRegistration : IEquatable<FilterRegistration>
     {
+        private static readonly LambdaEqualityHelper<FilterRegistration> equalityHelper =
+            new LambdaEqualityHelper<FilterRegistration>(x => x.Name);
+
         /// <summary>
         /// Maximum priority value a filter can have. A value of 200 will be run last on checkout and first on checkin.
         /// </summary>
-        public const int FilterPriorityMax = 200;
+        public const int PriorityMax = 200;
```
ethomson: 200 isn't the max priority, it's the default priority for things that use the `filter` attribute.  (Nor is `0` the minimum priority.)  If we're going to support arbitrary attributes, then we should support arbitrary priorities as well.

I think it's worth having a `DefaultPriority = 200` constant, though.

LibGit2Sharp/GlobalSettings.cs
```diff
@@ -192,47 +192,49 @@ public static IEnumerable<FilterRegistration> GetRegisteredFilters()
...
+        public static FilterRegistration<TFilter> RegisterFilter<TFilter>(string name, IEnumerable<string> attributes, int priority)
+            where TFilter : Filter, new()
         {
-            Ensure.ArgumentNotNull(filter, "filter");
-            if (priority < FilterRegistration.FilterPriorityMin || priority > FilterRegistration.FilterPriorityMax)
+            Ensure.ArgumentNotNull(name, "name");
+            Ensure.ArgumentNotNullOrEmptyEnumerable(attributes, "attributes");
+            if (priority < FilterRegistration.PriorityMin || priority > FilterRegistration.PriorityMax)
             {
                 throw new ArgumentOutOfRangeException("priority",
```
ethomson: This check isn't really correct - we should probably just defer to libgit2 to do the error checking on inputs.

LibGit2Sharp/GlobalSettings.cs
```diff
@@ -192,47 +192,49 @@ public static IEnumerable<FilterRegistration> GetRegisteredFilters()
...
             }
 
-            FilterRegistration registration = null;
+            FilterRegistration<TFilter> registration = null;
 
             lock (registeredFilters)
             {
                 // if the filter has already been registered 
-                if (registeredFilters.ContainsKey(filter))
+                if (registeredFilters.ContainsKey(name))
```
ethomson: Why do we duplicate the list of registered filters in LibGit2Sharp?  Why not defer to libgit2 for this?

LibGit2Sharp.Tests/FilterFixture.cs
```diff
@@ -392,62 +354,45 @@ private static void CreateAttributesFile(IRepository repo, string attributeEntry
...
-                }
+                CleanCallCount = 0;
+                InitializeCallCount = 0;
+                SmudgeCallCount = 0;
             }
 
-            protected override void Initialize()
+            public FakeFilter() { }
+
+            protected override void Apply(string path, string root, Stream input, Stream output, FilterMode mode)
```
ethomson: We shouldn't be providing the working directory root here.  Filters shouldn't be looking at the filesystem contents, they should be working on the streams.  So this is at best unnecessary and at worst suggests an antipattern to them.  If they (probably mistakenly) want to look at the working direcotry, then they should get that from the `Repository` themselves.

LibGit2Sharp/FilterRegistration.cs
```diff
@@ -52,35 +163,121 @@ internal FilterRegistration(Filter filter, int priority)
...
+        /// If non-NULL, the filter's `initialize` callback will be invoked right
+        /// before the first use of the filter, so you can defer expensive
+        /// initialization operations (in case libgit2 is being used in a way that doesn't need the filter).
+        /// </summary>
+        int InitializeCallback(IntPtr filterPointer)
+        {
+            int result = 0;
+            try
+            {
+                var methodInfo = typeof(TFilter).GetMethods()
```
ethomson: I don't understand why you're using reflection here...  This was very undiscoverable; I couldn't figure out how the `FakeFilter`'s `Initialize` method was getting called, but I knew it must be getting called because a unit test was validating that it was.

Even aside from the reflection being odd - this whole concept is a little odd.  I think that this really belongs at a higher level (`FilterRegistration`, for example) and not within the instance of a filter for a single file.

LibGit2Sharp.Tests/FilterFixture.cs
whoisj: This feature was added to support Git-LFS. Our friendly neighborhood GitHubbers are the requester of this feature.

LibGit2Sharp/FilterRegistration.cs
whoisj: Sure - I did not want to loose the functionality, but was not seeing a great way to restore it with a short-lived object. Adding it as a call back to `FilterRegistration` make sense. I'll make that change.

LibGit2Sharp/GlobalSettings.cs
whoisj: We used to, then the Garbage Collector would come through and clean up all the memory leaving the native library with invalid pointers. Once used would cause an AV which resulted in a NRE and the application crashing.

We keep a copy of the filter objects here to retain the callback entry-points necessary.

LibGit2Sharp/GlobalSettings.cs
whoisj: While that's fine, it does provide a less than ideal API for developers to use. Instead of having a concrete condition by which the inputs will be measured (and we can define in the xml comments), the values will be measured by the underlying library and a less meaningful exception thrown as a result.

Would it not be better to do the proper evaluation here and throw a more meaningful error?

LibGit2Sharp.Tests/FilterFixture.cs
ethomson: A) Why?
B) Why is using the `Repository` object to get this information not the right thing?

LibGit2Sharp/FilterRegistration.cs
whoisj: Ahh good to know! This is derived from the previous code and I just took what existed under the assumption that it was correct.

Is the minimum 0x80000000 and the max 0x7FFFFFFF?

LibGit2Sharp/GlobalSettings.cs
ethomson: There's two issues here, I think:

> Would it not be better to do the proper evaluation here ...

I don't know, that means keeping LibGit2Sharp's idea of the preconditions of something in sync with libgit2's, which seems like it's going to be hard to do.

> the values will be measured by the underlying library and a less meaningful exception thrown as a result.

If that's true, then we need to fix the error reporting.  Is this a theoretical concern or can you give me a concrete example?

In any case, there *are* no limitations here - things can be positive or negative and just within the range of a signed int, measured against (and ordered against) all the other filters that are registered.  (You explicitly want to be able to register a filter with a negative priority - the libgit2 filter tests illustrate this.)

LibGit2Sharp.Tests/FilterFixture.cs
whoisj: @ammeep would you mind explaining your needs for this feature?

Personally, I think having the information is valuable for two reasons:

  1. Assuming the filter will be a spawned process, it would be spawned with its working directory at the repository root
  2. Providing the information as part of the method call, avoid unnecessary `Repository` object allocations.

LibGit2Sharp/GlobalSettings.cs
whoisj: > In any case, there are no limitations here - things can be positive or negative and just within the range of a signed int, measured against (and ordered against) all the other filters that are registered. (You explicitly want to be able to register a filter with a negative priority - the libgit2 filter tests illustrate this.)

Ahh OK, now it makes more sense. There really is no valid check do then.

I'll remove the parameter check then.

whoisj:
/CC @nulltoken @ethomson 

I know this issue went stale, but I've come back to the project now and I'd like to revisit this.

Thanks!

PR #1144, opened 2015-07-08
Expose merge analysis

carlosmn:
This allows us to ask the library about what kind of actions
are available for merging into HEAD.

---

I'm not completely convinced by the overloads. Maybe we should expose `params Commit[]` and anything else just be a `params string[] committishes`.


carlosmn:
I'll reopen with the right target branch.

PR #1166, opened 2015-08-10
Add support revwalk range

xoofx:
Hi!
I have added support for [git_revwalk_push_range](https://libgit2.github.com/libgit2/#HEAD/group/revwalk/git_revwalk_push_range) on `CommitLog`/`CommitFilter`, as the current API doesn't support the behavior of the push_range (left part is not included, unlike the CommitFilter.Until).

I'm not particulary satisfied with the way the API is designed in CommitFilter. Not sure if we should better match the behavior of the underlying API instead. Maybe adding dedicated Push commands on `CommitFilter` would be more easier to map, but anyway, let me know.

I will add test once we agree on the design.
Thanks!

### DIFF
--- a/LibGit2Sharp/CommitFilter.cs
+++ b/LibGit2Sharp/CommitFilter.cs
@@ -108,10 +108,11 @@ namespace LibGit2Sharp
             if (obj is string)
             {
                 list.Add((string)obj);
-                return list;
+            } 
+            else if (obj is IEnumerable)
+            {
+                list.AddRange(((IEnumerable)obj).OfType<string>());
             }
-
-            list.AddRange(((IEnumerable)obj).OfType<string>());
             return list;
         }
     }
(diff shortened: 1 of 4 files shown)


xoofx:
> @nulltoken proposed in #1165 
> I thought a little more about your proposal. How would you feel about transforming your PR into a simpler static CommitFilter.BuildFrom() method which would return an instance of a CommitFilter?

If I had to redo it, I would rather prefer something that simply follow the native API. Remove CommitFilter, and provide a RevWalk object that exposes git_revwalk functions, but using it as an enumerator as well to make it more friendly to use (which is a bit merging `CommitFilter` and [`CommitEnumerator`](https://github.com/libgit2/libgit2sharp/blob/d11436877eeb5d632edf71a8b4ec0e45a5e7d4a0/LibGit2Sharp/CommitLog.cs#L132))

Typically with the current API, it is not possible to reuse a RevWalk object, to reset/modify its filters...etc.

Though, that's a design philosophy that doesn't match what is currently followed by LibGit2Sharp (which is technically more an higher level/abstraction layer than a wrapper), but that's a longer story :smile: 

So probably no need to completely refactor this. Whatever is done is fine for me, as long as we have somewhat exposed the `git_revwalk_push_range` functionality

PR #1173, opened 2015-08-18
The case for the removal of overloads which take an identity from the configuration

nulltoken:
These are meant to make it convenient to use the configured signature when the programmer wants to take it from the configuration.

But this makes these overloads very dependent on the configuration, something which is almost guaranteed to change between the developer's machine and the production (or even test) usage.

This causes code which was seemingly working fine to stop working once deployed (and this kind of environment dependence has bit us in libgit2sharp's tests as well) and it's letting the programmer assume that the environment is set up correctly instead of letting them make sure that it is or take steps to set it up correctly. This is not helped by `BuildSignature()` returning a fake signature if none is available in the configuration.

#1171 attempts to mitigate this problem by removing the fake signature. A programmer can now figure out whether the environment is set up correctly and set it up if that's the way they want to go (or conditionally provide their own signature). But it still lets the programmer make assumptions about the environment in which they are operating which can be confusing.

/cc @ckuhn203

### DIFF
--- a/LibGit2Sharp.Tests/CommitFixture.cs
+++ b/LibGit2Sharp.Tests/CommitFixture.cs
@@ -560,7 +560,9 @@ namespace LibGit2Sharp.Tests
 
                 Assert.Null(repo.Head[relativeFilepath]);
 
-                Commit commit = repo.Commit("Initial egotistic commit");
+                Signature signature = repo.Config.BuildSignature(DateTimeOffset.Now);
+
+                Commit commit = repo.Commit("Initial egotistic commit", signature, signature);
 
                 AssertBlobContent(repo.Head[relativeFilepath], "nulltoken\n");
                 AssertBlobContent(commit[relativeFilepath], "nulltoken\n");
--- a/LibGit2Sharp.Tests/CommitFixture.cs
+++ b/LibGit2Sharp.Tests/CommitFixture.cs
@@ -563,7 +563,7 @@ namespace LibGit2Sharp.Tests
         public void CanCommitWithSignatureFromConfig()
         {
             string repoPath = InitNewRepository();
-            string configPath = CreateConfigurationWithDummyUser(Constants.Signature);
+            string configPath = CreateConfigurationWithDummyUser(Constants.Identity);
             var options = new RepositoryOptions { GlobalConfigurationLocation = configPath };
 
             using (var repo = new Repository(repoPath, options))
@@ -586,7 +586,7 @@ namespace LibGit2Sharp.Tests
                 AssertBlobContent(repo.Head[relativeFilepath], "nulltoken\n");
                 AssertBlobContent(commit[relativeFilepath], "nulltoken\n");
 
-                AssertCommitSignaturesAre(commit, Constants.Signature);
+                AssertCommitIdentitiesAre(commit, Constants.Identity);
             }
         }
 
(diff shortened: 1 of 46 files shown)


nulltoken:
Transmuted this issue into a PR with a proposal. Thoughts?

ckuhn203:
As a consumer of the API, completely removing the `Commit(message)` overload is severely inconvenient. Passing a signature along with every commit is kind of clunky.

carlosmn:
You're still free to assume that `BuildSignature()` is going to find a signature, or create an extension method in your code which assumes your environment is always going to be set up just right. But that'd be an assumption *you* make about *your* environment, not something the library encourages.

ckuhn203:
All I'm saying is that it's not very elegant. I understand why it's needed, but it also doesn't make for a very nice API. 

Perhaps a property added to the `Repository` object? 

    using (var repo = New Repository)
    {
        repo.CurrentSignature = new Signature("rubberduck", "rubberduck@Foobar.com")

       //... do some things

       repo.Commit("message")
    }

I'm not a huge fan of that idea, because then Commit would have to check the property and throw an invalid operation exception if it was null. It makes assumptions about the state of the object that may not be true. Just thinking out loud about trying to find a more elegant solution.

dahlbyk:
@ckuhn203 brings up a good point. Even if we would consider it a best practice to explicitly specify an author, a `git commit` with name/email unset does exactly what we do now.

carlosmn:
Beware of trying to implement the git tool's UI. It can change without much notice, and this bit specifically has done so a few times in the past as it's meant for humans, not code.

>`git commit` with name/email unset does exactly what we do now.

If there is absolutely nothing configured, git will continue and create a commit with whatever it can find out of the current machine name and user account, which is going to be the wrong thing more often than not, but it errs on the side of making progress which is presumably better for the first steps (it does also tell you how to set the configuration). This does however mean that many people end up wanting to do a filter-branch of the first commits because their name isn't right on there.

What git does here is policy at the application layer, and what happens depends on it being a command-line application. If I were using a GUI without the user/email configuration set up, I would expect it to prompt me for the information. If I'm using some automation tool which performs a commit server-side, I would expect it to ask who to impersonate or use its name, as that information is much more useful than what system account it happens to run under. If it's a client-side automation tool, it's probably OK for it to use whatever is in the config, but I would expect it to be configurable.

All of these decisions are down to what the tool does and in what context is runs. We cannot pretend to know what is right for the consumers of the library. I also don't feel particularly compelled to make it convenient not to care about what information goes into a rather visible part of immutable history data. It also goes against making the right thing easer than the wrong thing. You can still make it completely dependent on the configuration by passing the config-generated signature, but it's at least as easy to put in `new Signature("my tool", "my-tool@example.com")`. Even better if this also goes with an explanation in the `Commit()` param that this is the signature for the current user or an identifier for the tool.

ckuhn203:
You've convinced me. We'll just need to update the wiki if/when this is implemented. I can see it causing a lot of confusion if it's not part of the Quick Start.

carlosmn:
Ah yes, we have one of those on this repo. Yes, the quick start would be a great place to say something like "the username from the config, or ask the user if you're committing on behalf of a user; if you're writing a tool, it's a good idea to identify it here so you know where the change came from".

dahlbyk:
:+1:

PR #1178, opened 2015-08-21
[RFC] Certs and Creds from a managed smart subtransport

carlosmn:
The current implementation does not ask for credentials or confirm server certs via the libgit2 channel; the implementation we know of wants to bypass libgit2 for these things. Which is fine, but if we want a managed implementation to behave like the built-in ones, we also need to let the subtransport ask for them.

This is an attempt at such an implementation. Basically nothing is final, but I'd like to know what we consider best to deal with return codes coming from the callbacks. We might be able to punt and say that we only support whatever libgit2sharp itself would return, so maybe we don't have to deal with `PASSTHROUGH` if we never generate it.

This depends on updated libgit2 binaries.


carlosmn:
I'm merging this since it's all new functionality so existing code will still work, and we can still try to figure out if something isn't quite right with the code in mainline.

PR #1193, opened 2015-09-23
LocalDev build scripts to reference locally-built libgit2

jeffhostetler:
These changes introduce a "LocalDev" version of the build/solution/project files that support using a locally-built debug version of LibGit2 while maintaining the overall structure of the existing NuGet-based scripts.  The goal here is to make it easier for a developer to work and debug at all 3 levels (App, LG2#, and LG2).

With these changes you can easily build and run the unit tests from the command line (using the new build.libgit2sharp.LocalDev.cmd script) and from VS (using the new LibGit2Sharp.LocalDev.sln).

Full details are provided in the new ./LocalDev/README.txt.

TODO I have only built this on Windows. I don't know if the other platforms need or have interest in something like this or not.


bording:
@nulltoken will have to make the final call on this, but this seems to be reintroducing a lot of stuff that was intentionally moved to the NativeBinaries repo. Putting this stuff back here now introduces a good bit of duplication that that would need to be maintained in two places.

ethomson:
Well, perhaps, but the current solution has created a number of problems that need solving.  Namely:

1. I want to use LibGit2Sharp without the public nuget packages.  I've reengineered my build system to emit libgit2 in a nuget package in the same way as LibGit2Sharp expects, which was obnoxious, but it's done.  Now what do I do with it?  See https://github.com/libgit2/libgit2sharp/pull/1181.

2. If I'm hacking on libgit2, I need a tight loop from code -> build -> use in LibGit2Sharp.  I absolutely don't want to build a nuget package to do it.  The best guidance seems to be to copy the new libgit2 into the nuget package's restored location.  This is the problem that this PR seems to be addressing.

I had pondered if (2) were also solveable with an environment variable or something pointing to the path to load the binary from.

Therzok:
> I had pondered if (2) were also solveable with an environment variable or something pointing to the path to load the binary from.

Probe env var and fallback to nuget? :heart:

ethomson:
> Probe env var and fallback to nuget? :heart:

Yeah - agreed that any solution should be to fall back to nuget so that the general getting started case remains nice and easy.

jeffhostetler:
Yeah, my original issue is that I need to hack on and/or single-step thru libgit2 as I'm working on a C# application.

And, yes there is way too much duplication of the various .sln and .csproj files doing it this way.

Let me take a look at the env var approach (now that I've figured out how the existing build scripts work).

jeffhostetler:
Abandoning this effort in favor of #1194.

PR #1207, opened 2015-10-22
Fluent PackBuilder

dahlbyk:
Extracted from https://github.com/libgit2/libgit2sharp/pull/1205

Is supporting `pb => pb.Add(obj1).Add(obj2)` good or useful?

### DIFF
--- a/LibGit2Sharp/PackBuilder.cs
+++ b/LibGit2Sharp/PackBuilder.cs
@@ -30,11 +30,13 @@ namespace LibGit2Sharp
         /// </summary>
         /// <param name="gitObject">The object to be inserted.</param>
         /// <exception cref="ArgumentNullException">if the gitObject is null</exception>
-        public void Add<T>(T gitObject) where T : GitObject
+        public PackBuilder Add<T>(T gitObject) where T : GitObject
         {
             Ensure.ArgumentNotNull(gitObject, "gitObject");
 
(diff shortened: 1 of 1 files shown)


ethomson:
Do you have a use case in mind?  My initial thought is that you would almost always do this in a loop and a fluent API wouldn't be particularly helpful.

dahlbyk:
> My initial thought is that you would almost always do this in a loop and a fluent API wouldn't be particularly helpful.

Having actually [written code using the API](https://github.com/libgit2/libgit2sharp/pull/1183#issuecomment-150276224), I tend to agree. In fact, `IEnumerable<>` overloads are probably worth having.

Of course, supporting chaining is trivial to implement with zero downside that I can see.

```diff
         internal void AddingObjectsTestDelegate(IRepository repo, PackBuilder builder)
         {
-            foreach (Branch branch in repo.Branches)
-            {
-                foreach (Commit commit in branch.Commits)
-                {
-                    builder.AddRecursively(commit);
-                }
-            }
-
-            foreach (Tag tag in repo.Tags)
-            {
-                builder.Add(tag.Target);
-            }
+            return packBuilder.AddRecursively(repo.Branches.SelectMany(b => b.Commits))
+                              .Add(repo.Tags.Select(t => t.Target));
         }
```

PR #1223, opened 2015-10-31
Run mono with --debug

carlosmn:
This makes mono load debug information from the assembly so we get e.g. line numbers in stack traces.


Therzok:
:+1:

PR #1230, opened 2015-11-04
README.md: fix imdisk.exe syntax

kevin-david:
I didn't actually run imdisk.exe last time, but pasting the command results in an error:
`Created device 2: X: -> Image in memory`
`Error defining drive letter: The handle is invalid.`

Removing the `:` fixes this error.


okletsgo:
same to me. after windwos 10 th2.

PR #1232, opened 2015-11-12
SubmoduleCollection.Add() using clone

JasonGoemaat:
I see two other pull requests for the same feature ([#482](https://github.com/libgit2/libgit2sharp/pull/482) and [#684](https://github.com/libgit2/libgit2sharp/pull/684)) that haven't been updated in 18 months, so I thought I would give it a try since I need this functionality for a project to convert from another source control system.

SubmoduleCollection.Add(url, subdir) acts like the command line 'git submodule add url subdir'.
It calls git_submodule_add_setup, then deletes the directory created by that and clones
into the same directory and calls git_submodule_add_finalize.  The end result is that the
.gitmodules file and the 'file' for the submodule directory are staged and ready to commit.

### DIFF
--- a/LibGit2Sharp.Tests/SubmoduleFixture.cs
+++ b/LibGit2Sharp.Tests/SubmoduleFixture.cs
@@ -267,6 +267,68 @@ namespace LibGit2Sharp.Tests
             }
         }
 
+
+        [Fact]
+        public void CanAddSubmoduleWithManualClone()
+        {
+            string configPath = CreateConfigurationWithDummyUser(Constants.Identity);
+            var options = new RepositoryOptions { GlobalConfigurationLocation = configPath };
+
+            //var path = SandboxSubmoduleTestRepo();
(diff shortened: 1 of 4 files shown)


carlosmn:
This functionality should not live in `SubmoduleCollection`. It'd be a good candidate for `Commands.SubmoduleAdd` or `Commands.Submdoule.Add` where it would be appropriate to emulate a high-level git command.

jwieneke:
Did this ever get into a release?
I'm looking exactly for something line this

> SubmoduleCollection.Add(url, subdir) acts like the command line 'git submodule add url subdir'.
> It calls git_submodule_add_setup, then deletes the directory created by that and clones into the same directory and calls git_submodule_add_finalize.  The end result is that the .gitmodules file and the 'file' for the submodule directory are staged and ready to commit.

There is no `Submodules.Add()`. I can't figure out if this is placed somewhere else like mentioned in the last comment.

PR #1240, opened 2015-12-01
Replace Global Filter Object with a Factory Model

whoisj:
Changed Filter to a short lived, per-file entity.

Standardized the exposed virtual entry-points.

Combined Clean and Smudge entry-points.

Moved all long-term constructs into `FilterRegister`.

Removed the superfluous `FilterAttributeEntry`.

Refactored to only support single attribute filters.

Added new exception type for easier identification of errors during odb <-> workdir operations.

Fixed up tests to use new model.

Sanity checked and rationalized substitution cipher filter tests.


carlosmn:
It looks like there's overlap with #1260 which we didn't realise when merging that. Is this (or parts of this) still needed?

whoisj:
> It looks like there's overlap with #1260 which we didn't realise when merging that. Is this (or parts of this) still needed?

They do overlap when considering the ability to run multiple streams simultaneously, but the semantics presented here are "cleaner" imo. Additionally, this version only keeps the registration itself persistent and invokes a new instance per invocation of the filter.

PR #1243, opened 2015-12-08
Correct the deprecation path for MergeConflictException

carlosmn:
We want users to be able to have the old name in their code and catch the
new type which we throw. To enable this we must switch the inheritance for
the old vs new type.

### DIFF
--- a/LibGit2Sharp.Tests/CheckoutFixture.cs
+++ b/LibGit2Sharp.Tests/CheckoutFixture.cs
@@ -1029,6 +1029,23 @@ namespace LibGit2Sharp.Tests
             }
         }
 
+        [Fact]
+        public void CanCatchDeprecatedException()
+        {
+            bool caught = false;
+
+            try
+            {
+                throw new CheckoutConflictException();
+            }
+            catch (MergeConflictException)
+            {
+                caught = true;
+            }
+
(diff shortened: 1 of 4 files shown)


LibGit2Sharp.Tests/CheckoutFixture.cs
```diff
@@ -1029,6 +1029,23 @@ public void CanCheckoutPathFromCurrentBranch(string fileName)
             }
         }
 
+        [Fact]
+        public void CanCatchDeprecatedException()
```
jamill: I don't think we necessarily need to assert this behavior.

jamill:
Looks good (I don't think we need to assert the ability to catch obsolete exceptions, though)

carlosmn:
The test was my way of figuring out what would work and I just left it there since it should help us avoid regressions till the release. The plan was to remove it just after the release when we delete the old name anyway, but sure, we don't have to include it here.

jamill:
That works as well! Either way seems fine. Thanks!
________________________________

nulltoken:
:+1: for me.

@whoisj any comment?

whoisj:
I think this should be fine. Does seem to make more sense this way around.

PR #1248, opened 2015-12-11
Explore moving the command-simulating methods to their own namespace

carlosmn:
A lot of the methods on Repository are actually trying to replicate git
commands, which makes it harder to see what's the operation vs command.

Move them to their own namespace to make it clear what they're trying to
achieve. We start with a few forms of the git-checkout command.

---

This shows a possible implementation of this. The `Checkout()` methods on Repository would still need to get deprecated, but they depend on each other in weird ways so we can figure that out if and when we decide to go forward with it.

The library itself should also likely get a similar treatment, but it's a lot more annoying to do anything in C, so we start here.

/cc @nulltoken @jamill @whoisj

### DIFF
--- a/LibGit2Sharp/Commands/Checkout.cs
+++ b/LibGit2Sharp/Commands/Checkout.cs
@@ -1,0 +1,100 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace LibGit2Sharp.Commands
+{
+    enum CheckoutMode
+    {
+        DetachHead,
+        CheckoutTree,
+    }
+
+    class Checkout
+    {
+        readonly Repository repo;
+        readonly Commit commit;
+        readonly Tree tree;
+        readonly IEnumerable<string> paths;
+        readonly CheckoutOptions options;
+        readonly CheckoutMode mode;
+
+        /// <summary>
+        /// Checkout a commit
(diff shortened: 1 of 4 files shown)


whoisj:
This is very nice! The only feature missing from the class is progress reporting! :smile:

carlosmn:
Progress reporting is going to be interesting if we want to use the same output as git. We can ask the user for a `Stream` to treat as stdout, but how would such a stream react to CR and clearing the line for the updates in the progress output? I suppose it's not worse than GUI apps that wrap `git.exe` and show `^K` at the end of lines.

whoisj:
@carlosmn  that is one option, another option is to return a streaming data structure which contains all of the relevant information. Should not be difficult to produce a structure which supports lines, colors, etc.

tamlin-mike:
>  We can ask the user for a Stream to treat as stdout

Just a brainstorming idea, but could a callback (delegate) work for progress reporting?
Would still require all the info (f.ex. colors), though. But two ideas reaching the same conclusion "this data is required" suggests it may be on the right track (a container for progress updates - even if only a specialization of a generic ColorableMessage or whatever).

It would fit quite well into GUI's I think, while a CUI implementation would be "obvious".

carlosmn:
Superseded by #1294

PR #1255, opened 2016-01-26
embed native libraries

reeflog:
and mark them as developement only for the nuget pack generation

### DIFF
--- a/LibGit2Sharp/Core/NativeMethods.cs
+++ b/LibGit2Sharp/Core/NativeMethods.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Globalization;
 using System.IO;
+using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.ConstrainedExecution;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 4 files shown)


reeflog:
This change would need a bit of clean-up around the deletion of files but I wanted to have a feedback if it worth it to do it. Does it has any chance to be merged ?

bording:
This probably isn't a good idea for a couple of reasons.

For starters, it's not going to work as you currently have it because you've assumed that only Windows has native binaries. The NativeBinaries package also has linux and OSX binaries. By marking the dependency as developmentDependency, the linux and OSX binaries won't be around on disk at all.

ethomson:
LibGit2Sharp is *not* going to try to write a native library to a temp directory on disk and then dynamically load it.  This is a *massive* security risk, since temp directories are quite often not secured and it's quite trivial for an (unprivileged) attacker to quickly write a different native library that you will then load and execute.

If that's something that you want your application to do (and I strongly recommend that you *do not* do this) then you are welcome to.  You can lay that down on disk and then update the native settings location yourself to point to that location.

reeflog:
I was not convince myself of the solution but it was the quicker I found.
Thanks for the feedbacks and advices.

PR #1257, opened 2016-02-15
Update CHANGES for 0.22

carlosmn:


ethomson:
Coooooooool.  Let's add #1258 to this and then :shipit: .

PR #1260, opened 2016-02-17
Fix crash when running multiple filter smudge or clean operations concurrently

niik:
The current filter implementation in libgit2sharp stores the input/output stream pointers and more as instance variables on the `Filter` class. Unfortunately this makes it impossible to run multiple filter streams concurrently.

Due to the nature of the problem manifests itself in a few different ways, sometimes as an unmanaged crash due to invalid pointer access, sometimes as a managed exception. When run under the debugger you'll sometimes get notified that a callback has been collected.

```
CallbackOnCollectedDelegate occurred
Message: Managed Debugging Assistant 'CallbackOnCollectedDelegate' has detected a problem in 'Foo.Exe'.
Additional information: A callback was made on a garbage collected delegate of type 'LibGit2Sharp!LibGit2Sharp.Core.GitWriteStream+close_fn::Invoke'. This may cause application crashes, corruption and data loss. When passing delegates to unmanaged code, they must be kept alive by the managed application until it is guaranteed that they will never be called.
```

This PR changes the implementation to root the necessary references in a dictionary keyed by the stream pointer.

cc @carlosmn @ethomson

### DIFF
--- a/LibGit2Sharp.Tests/FilterFixture.cs
+++ b/LibGit2Sharp.Tests/FilterFixture.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.IO;
 using LibGit2Sharp.Tests.TestHelpers;
 using Xunit;
+using System.Threading.Tasks;
 
 namespace LibGit2Sharp.Tests
 {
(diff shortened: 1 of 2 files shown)


carlosmn:
Looks good, could you squash the commits after f7e50b2 into ti? We try to keep the history free of small commits fixing up those from before in a PR.

niik:
:koala:

carlosmn:
:bear:

PR #1271, opened 2016-03-04
Bump version number to 0.23 (prerelease)

ethomson:
Historically, we have been bumping the version number immediately after releasing, so that our new packages take the new (prerelease) version numbers.

Update to use 0.23.0 prerelease version numbers.


ethomson:
That's right.  Let's put the letters `:wq` at the end of our commit message.  Sigh.

PR #1278, opened 2016-03-15
Add a Peel method to GitObject

carlosmn:
This exposes what we've only had internally, letting you ask for e.g. a
tree from a tag.

Make use of this internally as well instead of the special-cased method
we had up to now.

---

I wanted to do something cute like attach the `GitObjectType` value to the type itself so I could do `T.GitObjectType` instead of performing a dictionary lookup, but C# won't let me have a per-class static field.

I wonder if having the dictionary is worth the overhead or a function with a bunch of `if`s would work out to be quicker (it's likely not a big deal either way).


ethomson:
Merged manually.

PR #1279, opened 2016-03-16
Expose the underlying reference for ReferenceWrapper

carlosmn:
This lets us get to the reference's information without having to ask
the wrapper object for the information.

/cc @niik

### DIFF
--- a/LibGit2Sharp/ReferenceWrapper.cs
+++ b/LibGit2Sharp/ReferenceWrapper.cs
@@ -16,7 +16,7 @@ namespace LibGit2Sharp
         /// The repository.
         /// </summary>
         protected readonly Repository repo;
-        protected readonly Reference reference;
+        private readonly Reference reference;
         private readonly Lazy<TObject> objectBuilder;
 
         private static readonly LambdaEqualityHelper<ReferenceWrapper<TObject>> equalityHelper =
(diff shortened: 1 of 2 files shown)


niik:
This looks great to me @carlosmn! Did you want to do the explicit direct reference resolve here as well?

carlosmn:
Not as part of this PR. I think that would involve some changes in what we accept, since we'd only want `DirectReference`s and change what `Repository.Head` returns to not be a branch and related things so we can avoid the type system thinking a detached HEAD is a branch.

niik:
Great :+1:

LibGit2Sharp/ReferenceWrapper.cs
```diff
@@ -16,7 +16,10 @@ public abstract class ReferenceWrapper<TObject> : IEquatable<ReferenceWrapper<TO
         /// The repository.
         /// </summary>
         protected readonly Repository repo;
-        protected readonly Reference reference;
+        /// <summary>
+        /// The underlying reference
+        /// </summary>
+        readonly Reference reference;
```
ethomson: I don't super love using default visibility here - it seems like we're explicit everywhere else.  Can we make this explicit here, too?  (Plus it saves me a google since I can never remember what the defaults are between C# and Java).

carlosmn: Fixed

PR #1280, opened 2016-03-17
Update to libgit2 de143ef

ethomson:
https://github.com/libgit2/libgit2/compare/785d8c4...de143ef

### DIFF
--- a/LibGit2Sharp/packages.config
+++ b/LibGit2Sharp/packages.config
@@ -1,4 +1,4 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="LibGit2Sharp.NativeBinaries" version="1.0.129" targetFramework="net4" allowedVersions="[1.0.129]" />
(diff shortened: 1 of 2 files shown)


LibGit2Sharp/packages.config
```diff
@@ -1,4 +1,4 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="LibGit2Sharp.NativeBinaries" version="1.0.129" targetFramework="net4" allowedVersions="[1.0.129]" />
+  <package id="LibGit2Sharp.NativeBinaries" version="1.0.132" targetFramework="net40" />
```
carlosmn: AFAICT we need to keep the `allowedVersions` field, or we risk nuget/VS using the wrong package

ethomson: Thanks, I did this upgrade in VS and apparently didn't do it right.  Fixed.

PR #1281, opened 2016-03-17
Lazy TreeChanges, move ownership of diff handle for diff results

niik:
This will let us perform lazy loading of diff details in `Patch`, `PatchStats` and `TreeChanges` just as we do with commits and some other classes.

In this initial PR I've only worked with `TreeChanges` which was where our performance concerns were. We need the fastest possible way to get the number of changed files between two trees.

In order for this to be possible I had to move ownership of the diff handle to the individual diff result classes. To keep this PR as small as possible I've kept the behavior of `Patch` and `PatchStats` such that while they're now disposable they eagerly load everything and dispose of the handle in their respective constructors.

cc @ethomson @carlosmn

### DIFF
--- a/LibGit2Sharp/TreeChanges.cs
+++ b/LibGit2Sharp/TreeChanges.cs
@@ -39,8 +39,7 @@ namespace LibGit2Sharp
         private IEnumerable<TreeEntryChanges> GetChangesOfKind(ChangeKind changeKind)
         {
             TreeEntryChanges entry;
-
-            for (int i = 0; i < count.Value; i++)
+            for (int i = 0; i < Count; i++)
             {
                 if (TryGetEntryWithChangeTypeAt(i, changeKind, out entry))
                 {
@@ -78,7 +77,7 @@ namespace LibGit2Sharp
         /// <returns>An <see cref="IEnumerator{T}"/> object that can be used to iterate through the collection.</returns>
         public virtual IEnumerator<TreeEntryChanges> GetEnumerator()
         {
-            for (int i = 0; i < count.Value; i++)
+            for (int i = 0; i < Count; i++)
             {
                 yield return GetEntryAt(i);
             }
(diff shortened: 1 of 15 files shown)


carlosmn:
Neat! My end-game would be to get rid of the three types and just let you query from a general `Diff` object but this is a nice improvement over what we have now.

LibGit2Sharp/TreeChanges.cs
```diff
@@ -52,21 +28,28 @@ protected TreeChanges()
 
         internal TreeChanges(DiffSafeHandle diff)
         {
-            Proxy.git_diff_foreach(diff, FileCallback, null, null);
+            this.diff = diff;
+            this.count = new Lazy<int>(() => Proxy.git_diff_num_deltas(diff));
```
Therzok: This is not safe. What if the internal diff changes between the first access to Count and the time of enumeration?

niik: @Therzok Oh I see. I can definitely remove the memoization here. I was (mistakenly apparently) under the impression that diffs were immutable once created. Out of curiosity, could you point me to a scenario where count would mutate?

carlosmn: When would a diff change its size once it's been computed?

Therzok: Oh, I just realised that the `DiffSafeHandle` is saved too. Discard my comment.

ethomson: A diff's size _could_ change during rename detection, but that doesn't seem like something that's going to happen here.  :)

niik: > A diff's size could change during rename detection, but that doesn't seem like something that's going to happen here. :)

Yeah, and at the point where this class gets passed the diff handle rename detection (if enabled) would already have occurred so we should be safe, right?

niik:
Look at that green CI at last. I think this is ready for review. This is for all intents and purposes a breaking API change since we now require callers of `Compare<T>` to dispose of the return values themselves (except not really for `Patch*`).

carlosmn:
:+1: from me, can you rebase?

niik:
> :+1: from me, can you rebase?

Done, I don't know what's up with CI but it looks unrelated to my stuff and there's a few other PRs that's failing for the same reason so I'm suspecting master is slightly borked

PR #1284, opened 2016-03-19
Eagerly dispose of Remotes

niik:
This is a continuation of #1249. Since Remotes are now disposable we should eagerly clean up their resources rather than wait for GC to call
the destructor.

cc @ethomson @carlosmn

### DIFF
--- a/LibGit2Sharp/BranchUpdater.cs
+++ b/LibGit2Sharp/BranchUpdater.cs
@@ -154,7 +154,7 @@ namespace LibGit2Sharp
                 if (!remoteName.Equals(".", StringComparison.Ordinal))
                 {
                     // Verify that remote exists.
-                    repo.Network.Remotes.RemoteForName(remoteName);
+                    using (repo.Network.Remotes.RemoteForName(remoteName)) { }
                 }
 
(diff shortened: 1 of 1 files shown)


niik:
See https://github.com/libgit2/libgit2sharp/pull/1249#issuecomment-198639995

niik:
There's also a load of tests which use `Remote` that I didn't touch here

carlosmn:
We could add

```diff
diff --git a/LibGit2Sharp/Remote.cs b/LibGit2Sharp/Remote.cs
index f92b4aa..a318707 100644
--- a/LibGit2Sharp/Remote.cs
+++ b/LibGit2Sharp/Remote.cs
@@ -39,6 +39,7 @@ internal Remote(RemoteSafeHandle handle, Repository repository)
             PushUrl = Proxy.git_remote_pushurl(handle);
             TagFetchMode = Proxy.git_remote_autotag(handle);
             refSpecs = new RefSpecCollection(this, handle);
+            repository.RegisterForCleanup(this);
         }
 
         ~Remote()
```

which makes us release the handle when the repository itself gets disposed.

PR #1285, opened 2016-03-21
Take a name instead of an instance to update a remote

carlosmn:
All the Remote instance does in RemoteUpdater is carry the name, as we
need it in order to perform the update. Reflect this in the class fields
as well as in the preferred way of calling it.

While we're dealing with the updater, make it perform its actions inside
a configuration transaction so the configuration does not change while
we're processing updates.


carlosmn:
This needs libgit2/libgit2#3703 since writing multivars under lock is buggy for silly reasons.

ethomson:
libgit2/libgit2#3703 landed in our natives, so this looks good to go!

PR #1286, opened 2016-03-21
Remove equality comparison from Remote

carlosmn:
There are no semantics for determining whether a remote is like another
one. The equality decision in libgit2sharp only looks at the name and
urls, ignoring any other number of fields which affect how a remote
behaves.

For these reaons, rip it out. Equality (in as much as it's valuable)
should be perform on whatever aspect the code is interested.

---

If there's a way to deprecate the implementation of an interface I'm all ears, but it looks like we might want to simply get rid of it.


whoisj:
:+1:

PR #1287, opened 2016-03-22
Add a RemoteName property to avoid lookups

niik:
Sometimes all you need is the name and right now there's no way of getting it short of `git_remote_lookup`. While going through a lookup guarantees that the remote exists that shouldn't be the concern of the `Branch`, it should just let the caller know what's been configured as it's remote.

cc @carlosmn We could/should ditch the `Remote` property here IMO seeing how it's disposable now and have callers to go through `Repository.Remotes`

cc @ethomson

### DIFF
--- a/LibGit2Sharp/Branch.cs
+++ b/LibGit2Sharp/Branch.cs
@@ -182,6 +182,7 @@ namespace LibGit2Sharp
         ///   the tracked branch.
         /// </para>
         /// </summary>
+        [Obsolete("This property is deprecated. Use Repository.Network.Remotes[] using the RemoteName property")]
         public virtual Remote Remote
         {
             get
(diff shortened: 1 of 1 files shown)


LibGit2Sharp/Branch.cs
```diff
@@ -154,6 +154,26 @@ public virtual string UpstreamBranchCanonicalName
         }
 
         /// <summary>
+        /// Get the name of the remote for the branch.
+        /// <para>
+        ///   If this is a local branch, this will return the configured
+        ///   <see cref="Remote"/> to fetch from and push to. If this is a
+        ///   remote-tracking branch, this will return the remote containing
+        ///   the tracked branch. If there no tracking information this will
+        ///   return null.
```
carlosmn: This should not say it returns a remote but its name.

carlosmn:
Let's deprecate the `Remote` property with a note saying to use `RemoteName`.

niik:
> Let's deprecate the Remote property with a note saying to use RemoteName.

Done

PR #1288, opened 2016-03-22
Make Pull and Fetch commands

carlosmn:
These want to behave like commands, so let's put them in a namespace that says as much.

This is similar to #1248 but I'm touching the network area so I wanted to send this one as well, so we can discuss the design with more data points.

In particular, should these be classes? They don't really do anything that needs a class. A static method in a static class might work out better for keeping namespacing but not forcing undue instantiations.

These commands take a string instead of having `Remote` and `string` variants since that's what a command would take and all the `Remote` instances were doing was carry the name.

### DIFF
--- a/LibGit2Sharp.Tests/FetchFixture.cs
+++ b/LibGit2Sharp.Tests/FetchFixture.cs
@@ -22,7 +22,7 @@ namespace LibGit2Sharp.Tests
 
             using (var repo = new Repository(path))
             {
-                Remote remote = repo.Network.Remotes.Add(remoteName, url);
+                repo.Network.Remotes.Add(remoteName, url);
 
                 // Set up structures for the expected results
                 // and verifying the RemoteUpdateTips callback.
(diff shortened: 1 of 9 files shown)


txdv:
Maybe put them in extension methods of Network?

The class approach would kinda work if you would make the arguments accessible:

```csharp
        private readonly Repository repository;
        private readonly Signature merger;
        private readonly PullOptions options;
```

LibGit2Sharp.Tests/FetchFixture.cs
```diff
@@ -22,7 +22,7 @@ public void CanFetchIntoAnEmptyRepository(string url)
 
             using (var repo = new Repository(path))
             {
-                Remote remote = repo.Network.Remotes.Add(remoteName, url);
+                repo.Network.Remotes.Add(remoteName, url);
```
Therzok: Wouldn't this cause an issue, since `Remote` is now `IDisposable`?

I know this is a test method, but we should show the _right thing_™

carlosmn: It's in the same situation as before. We dispose of the remotes on repository disposal.

carlosmn:
Making this extension methods is the opposite direction to what this is trying to achieve. It's an attempt at making clear what is an operation and what is a method that's trying to replicate what you'd call on the command line.

ethomson:
Yep.  I like this.

> In particular, should these be classes?

Nah.  Overkill, I think.

ethomson:
LET THE COMMANDIFICATION BEGIN.

PR #1291, opened 2016-03-23
Move Stage and Remove to commands

carlosmn:
Similarly to the other PRs, let's move away from having these in the repository. Making this external also means it'll work for more situations than it used to since now it expects to work against an `IRepository` instead of from inside a `Repository`.

This also has us make the `Index` not write itself to disk on every single operation, so we end up doing less I/O while we build up the index we do want to keep.

I think this would be something you use @niik so do you have thoughts?


ethomson:
Sorry, after merging #1288, there's conflicts here now.  :/

PR #1292, opened 2016-03-29
Ignore .vs/ folder

AArnott:


ethomson:
Surprised we didn't already have this.  Thanks for adding it.

PR #1300, opened 2016-04-12
Use Travis' csharp language setup

### LINKED ISSUE #1032: Use LEAKS defines on Mono and use Mono 4.
We should upgrade the mono used on travis to use Mono 4 as soon as it is avaiable. I will update this issue when the mono-project website reflects this info.

### LINKED ISSUE #1082: Leverage Travis csharp language

carlosmn:
Use the configuration Travis already knows about to reduce how much
custom code we're running.

As it's now easy to do, test on both 3.12 and 4.2 versions of mono. This
also has the advantage of making us test on both systems with the same
version.

This fixes #1032, fixes #1082, supersedes #1086, supersedes #1129


carlosmn:
This seems like it's working. We build with 3.12 so we support those who can't upgrade for whatever reason, but also with 4.2 so we build with something recent. It also unifies which versions we test on which OS, which was different between Ubuntu and OS X. @ethomson thoughts?

ethomson:
This is really delightful.  Thanks for tackling this.  And thanks @FeodorFitsner for making this so easy.

PR #1305, opened 2016-04-19
Expose merge analysis

carlosmn:
This allows us to ask the library about what kind of actions
are available for merging into HEAD.

---

I'm still not quite sure what the overloads should be. Maybe leave as-is and think about exposing annotated commits as a type.

### DIFF
--- a/LibGit2Sharp.Tests/MergeFixture.cs
+++ b/LibGit2Sharp.Tests/MergeFixture.cs
@@ -874,6 +874,72 @@ namespace LibGit2Sharp.Tests
             return repository.Commit("New commit", Constants.Signature, Constants.Signature);
         }
 
+
+        [Theory]
+        [InlineData("master", MergeAnalysis.Unborn | MergeAnalysis.FastForward, true)]
+        [InlineData("fast_forward", MergeAnalysis.Normal | MergeAnalysis.FastForward, false)]
+        [InlineData("conflicts", MergeAnalysis.Normal, false)]
+
+        public void CanAnalyzeMerge(string branchName, MergeAnalysis analysis, bool makeUnborn)
+        {
+            string path = SandboxMergeTestRepo();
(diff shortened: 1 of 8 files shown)


ethomson:
Yeah, I worry about having these two overloads.  Because what happens when you want to octopus merge two commits and a ref?  (Well, this is entirely theoretical, since libgit2 doesn't even do octopus merging, but you know.)

I wonder if `Commit` and `Reference` should implement `IMergeable` or something?

carlosmn:
You can peel the `Reference` to a commit, that shouldn't be much of an issue. They _could_ implement `IAnnotatable` or something since what we actually want is an annotated commit and these are just convenience methods for that.

ethomson:
> You can peel the Reference to a commit, that shouldn't be much of an issue. 

Sure - I guess my thinking here is twofold:  first, it would be nice if we didn't make users do the peeling.  :)  And second, `Merge` will behave differently when handed a commit versus an actual branch.  _Merge analysis_, of course, will not but I'm wondering if we can make sure these APIs are nice and similar.  At the moment we have:

        MergeResult Merge(Commit commit, Signature merger, MergeOptions options);
        MergeResult Merge(Branch branch, Signature merger, MergeOptions options);
        MergeResult Merge(string committish, Signature merger, MergeOptions options);

And it would definitely be nice to get rid of these overloads and have a single `Merge` that takes the same type of commits/references/etc that the analysis method takes...

carlosmn:
Sure, we can accept a `IAnnotateCommit` that knows how to get the actual handle we want from a branch/ref/revparse. It's a shame you can't implement it for `string` but we can have two overloads, one for rev-parse strings and one for objects.

carlosmn:
Updated with more interfaces, yay. The CI error seems to come from us assuming interfaces are only ever there for mocking, which seems like a bad idea.

PR #603, opened 2014-01-10
Posix style paths for object database elements

clementbellot:
See issue #596.

PR #1307, opened 2016-04-20
Don't convert to Windows-style paths when dealing with Git data

carlosmn:
I've ported the tests from #603 on top of master to check what we still had to do, but they pass on my system.

So we should probably add them to make sure they don't break, but AFAICT these situations are already handled.

### DIFF
--- a/LibGit2Sharp.Tests/DiffTreeToTreeFixture.cs
+++ b/LibGit2Sharp.Tests/DiffTreeToTreeFixture.cs
@@ -10,7 +10,7 @@ namespace LibGit2Sharp.Tests
 {
     public class DiffTreeToTreeFixture : BaseFixture
     {
-        private static readonly string subBranchFilePath = Path.Combine("1", "branch_file.txt");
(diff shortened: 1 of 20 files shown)


carlosmn:
Oh right, I forgot this would only show up on Windows where the path combination char is problematic.

carlosmn:
This rabbit hole goes quite deep. Turns out we convert to "native" paths rather often. I think anything we return should just be whatever libgit2 returns. Otherwise you're going to have trouble figuring out what to expect on what platform. And in any case `Path.Combine()` won't work for datastructures if you want a filename that Win32 happens not to like.

ethomson:
Yuuuuup.  Thanks for doing this.  Are there other places that you know we're conflating paths and strings, or is this ready to merge?

carlosmn:
>  Are there other places that you know we're conflating paths and strings

¯\_(ツ)_/¯ I think we're probably conflating them much more often than we notice. I'll go over `NativeMethods` and check if we're using `StrictFilePathMarshaller` where we should be passing whatever string the user gave us.

PR #1310, opened 2016-04-26
Remove FollowFilter for add an override of QueryBy

### LINKED ISSUE #1276: reinforcement for Commits.QueryBy
Can we introduce the override `QueryBy(string path, CommitFilter filter)` ?

usage like
```
var commits = repo.Commits.QueryBy(path, new CommitFilter { IncludeReachableFrom = commit })
```

Aimeast:
`FollowFilter` is a duplicate with `CommitFilter`. We only need ensure the `SortStrategies` has checked before create `FileHistory`.

Visual Studio tests passed except `NoPublicTypesUnderLibGit2SharpCoreNamespace`

to close #1276

### DIFF
--- a/LibGit2Sharp.Tests/FileHistoryFixture.cs
+++ b/LibGit2Sharp.Tests/FileHistoryFixture.cs
@@ -98,7 +98,7 @@ namespace LibGit2Sharp.Tests
 
                 // Test --date-order.
                 var timeHistory = repo.Commits.QueryBy(path,
-                    new FollowFilter { SortBy = CommitSortStrategies.Time });
+                    new CommitFilter { SortBy = CommitSortStrategies.Time });
                 var timeCommits = new List<Commit>
                 {
                     master10, // master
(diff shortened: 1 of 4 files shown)


carlosmn:
We can't remove a type without deprecating it for a release.

Aimeast:
updated. mark FollowFilter as obsolete

Aimeast:
Can I kindly ask what is your plan for merge this. My project need to use this feature. Could you please tell me? Thank you!

ethomson:
Hi @Aimeast, thanks for the ping.

I think that in order to deprecate this politely, we want to warn people about `FollowFilter`'s deprecation (which you've done) but also allow it to work until its removed in the next release.

We could do that by either having a second overloaded method that takes `FollowFilter` and converts the param to a `CommitFilter`s.  Alternately, we could simply make `FollowFilter` extend `CommitFilter`.  That way people who were creating a `FollowFilter` would be able to pass it to the function that takes a `CommitFilter`.

Does that also sound like it will work?  If so, could you make that change and then I will merge it.  Thanks!

Aimeast:
@ethomson thanks for your quickly response.

I have updated PR. Added a overload to `CommitLog` and `IQueryableCommitLog`

ethomson:
Neat, thanks for fixing this!  😁

Issue #1312, opened 2016-05-05
Test failure: NoPublicTypesUnderLibGit2SharpCoreNamespace


AArnott:
This test fails in the master branch (as of 1fb7fc5b1d26c893e085c2782d5ee483d56c5a18)

```
Test Name:	LibGit2Sharp.Tests.MetaFixture.NoPublicTypesUnderLibGit2SharpCoreNamespace
Test FullName:	LibGit2Sharp.Tests.MetaFixture.NoPublicTypesUnderLibGit2SharpCoreNamespace
Test Source:	C:\Users\andarno\git\libgit2sharp\LibGit2Sharp.Tests\MetaFixture.cs : line 298
Test Outcome:	Failed
Test Duration:	0:00:00.003

Result StackTrace:	at LibGit2Sharp.Tests.MetaFixture.NoPublicTypesUnderLibGit2SharpCoreNamespace() in C:\Users\andarno\git\libgit2sharp\LibGit2Sharp.Tests\MetaFixture.cs:line 324
Result Message:	Public type 'LibGit2Sharp.Core.LeaksContainer' under the 'LibGit2Sharp.Core' namespace.
```

carlosmn:
This happens because the compilation defines `LEAKS_IDENTIFYING` which makes us create this flag, but the tests got compiled by VS itself which did not, thus not making us include `LeaksContainer` in the list of classes that are OK to export.

We should probably get rid of this long-term as we'd actually want to expose things in core.

carlosmn:
Does #1317 fix it for you?

ethomson:
I think #1317 should fix this - @AArnott please feel free to reopen if it doesn't.

PR #1316, opened 2016-05-05
Libgit2Object: make visibility internal

ethomson:
The unit tests are failing because the leak checker is public.  See #1312.


ethomson:
It seems that the tests look at this, so this isn't going to without an internalsvisibleto or some other trickery.  @niik or @carlosmn do you have clever ideas?

carlosmn:
It's a harmless error, but if we're gonna keep this test, I think we should turn to reflection to figure out if `LeaksContainer` was defined and if so, add it to the list of public objects to ignore.

PR #1317, opened 2016-05-06
Check for LeaksContainer via a string

carlosmn:
This means we don't need to define LEAKS_IDENTIFYING in the tests, which
we don't control, as the IDE might rebuild if something has changed.


ethomson:
Works for me!

PR #1323, opened 2016-05-31
Add Cherry-pick and Revert functionality to ObjectDatabase

jkoritzinsky:
This PR adds support for cherry-picking and reverting commits using the `*_commit` functions from libgit2 to allow cherry picking and reverting on a bare repository.


LibGit2Sharp/RevertTreeResult.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// The resulting conflicts from the cherry-pick.
+        /// </summary>
+        public virtual IEnumerable<Conflict> Conflicts { get; private set; }
+
+        /// <summary>
+        /// The status of the cherry pick.
+        /// </summary>
+        public virtual MergeTreeStatus Status { get; private set; }
+    }
+}
```
jamill: nit: can we end files with a trailing newline

LibGit2Sharp/RevertTreeResult.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System.Collections.Generic;
+
+namespace LibGit2Sharp
+{
+    public class RevertTreeResult
```
jamill: Can we include a class comment

LibGit2Sharp/RevertTreeResult.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        internal RevertTreeResult(IEnumerable<Conflict> conflicts)
+        {
+            this.Conflicts = conflicts;
+            this.Status = MergeTreeStatus.Conflicts;
+        }
+
+        /// <summary>
+        /// The resulting tree of the cherry-pick.
```
jamill: I think these comments have some left over text from the cherry-pick version :) Would you mind just updating them to refer to the revert action?

LibGit2Sharp/ObjectDatabase.cs
```diff
@@ -806,5 +880,79 @@ private PackBuilderResults InternalPack(PackBuilderOptions options, Action<PackB
...
+                MergeFileFavorFlags = options.MergeFileFavor,
+                MergeTreeFlags = mergeFlags,
+                RenameThreshold = (uint)options.RenameThreshold,
+                TargetLimit = (uint)options.TargetLimit
+            };
+
+            bool earlyStop;
+
+            using (var ourHandle = Proxy.git_object_lookup(repo.Handle, ours.Id, GitObjectType.Commit))
+            using (var cherryPickCommitHandle = Proxy.git_object_lookup(repo.Handle, revertCommit.Id, GitObjectType.Commit))
```
jamill: leftover variable name from cherryPick?

LibGit2Sharp/ObjectDatabase.cs
```diff
@@ -542,6 +542,80 @@ public virtual HistoryDivergence CalculateHistoryDivergence(Commit one, Commit a
...
 
         /// <summary>
+        /// Performs a cherry-pick of <paramref name="cherryPickCommit"/> onto <paramref name="ours"/> commit.
+        /// </summary>
+        /// <param name="cherryPickCommit">The commit to cherry-pick.</param>
+        /// <param name="ours">The commit to cherry-pick onto.</param>
+        /// <param name="mainline">Which commit to consider the parent for the diff when cherry-picking a merge commit.</param>
+        /// <param name="options">The options for the merging in the cherry-pick operation.</param>
+        /// <returns>A result containing a <see cref="Tree"/> if the cherry-pick was successful and a list of <see cref="Conflict"/>s if it is not.</returns>
+        public virtual CherryPickTreeResult CherryPickCommit(Commit cherryPickCommit, Commit ours, int mainline, MergeOptions options)
```
jamill: Does `ours` work for cherry-pick? What would you think of `cherryPickOnto` instead of `ours`?

LibGit2Sharp/LibGit2Sharp.csproj
```diff
@@ -393,4 +395,4 @@
   </Target>
   -->
   <ItemGroup />
-</Project>
+</Project>
```
jamill: A bit harder for generated files - but instead of fighting about the trailing newline, could you go back and add it here?

LibGit2Sharp.Tests/RevertFixture.cs
```diff
@@ -458,5 +458,77 @@ public void RevertOrphanedBranchThrows()
...
+                Assert.Equal(MergeTreeStatus.Conflicts, result.Status);
+                Assert.Null(result.Tree);
+            }
+        }
+
+        [Fact]
+        public void CanRevertInObjectDatabase()
+        {
+            // The branch name to perform the revert on,
+            // and the file whose contents we expect to be reverted.
```
jamill: Is the comment about the file still relevant?

LibGit2Sharp.Tests/RevertFixture.cs
```diff
@@ -458,5 +458,77 @@ public void RevertOrphanedBranchThrows()
...
+            const string revertBranchName = "refs/heads/revert";
+
+            string path = SandboxRevertTestRepo();
+            using (var repo = new Repository(path))
+            {
+
+                
+                // Checkout the revert branch.
+                Branch branch = Commands.Checkout(repo, revertBranchName);
```
jamill: Do we need to checkout the branch in order to do this revert?

jamill:
I think it would be better if both `RevertTreeResult` and `CherryPickTreeResult` contained a consistent status property... Either both use the `MergeTreeStatus`, or both use `RevertTreeStatus` and `CherryPickTreeStatus` respectively... (I think I would opt for the former? as I think the set of possibilities are the same?)

jamill:
Thank you @jkoritzinsky for this very nice PR!

LibGit2Sharp.Tests/RevertFixture.cs
jkoritzinsky: We need to either checkout and do the revert on the checked out branch or generate a commit from the resulting tree after the first revert and apply that before the second revert. The code is much cleaner to check out the branch to perform the first revert on.

ethomson:
This looks really good to me.  One question:  do we need separate `*TreeResult` classes?  It seems like either we should use a single class for this, or that these should inherit from a base class.

jkoritzinsky:
We can use a single class, but I am having trouble coming up with a good name. Do you have any suggestions?

ethomson:
I, personally, would be okay with using `MergeTreeResult`, either completely, or just as the base class.  But I can understand that this is perhaps a bit of a leaky abstraction.  (It's not necessarily obvious that a cherry-pick or a revert is really a funny merge.)

jamill:
>  I, personally, would be okay with using  MergeTreeResult 

👍

jkoritzinsky:
I also like `MergeTreeResult`. I'll go change it to that.

ethomson:
Great!  Can you squash this and I'll merge it?  Thanks!

jkoritzinsky:
Squashed!

PR #1328, opened 2016-06-18
Include Ignored Options

paulgmiller:
We have some repos that produce thousands of ignored files so retrieve status takes 20 minute on them if we include ignore files vs a couple of seconds if we don't. There fore we were hoping we could make including ignored files optional.

### DIFF
--- a/LibGit2Sharp/RepositoryStatus.cs
+++ b/LibGit2Sharp/RepositoryStatus.cs
@@ -79,10 +79,14 @@ namespace LibGit2Sharp
                 Version = 1,
                 Show = (GitStatusShow)options.Show,
                 Flags =
-                    GitStatusOptionFlags.IncludeIgnored |
                     GitStatusOptionFlags.IncludeUntracked |
                     GitStatusOptionFlags.RecurseUntrackedDirs,
             };
+
+            if (options.IncludeIgnored)
+            {
+                coreOptions.Flags |= GitStatusOptionFlags.IncludeIgnored;
+            }
 
             if (options.DetectRenamesInIndex)
             {
--- a/LibGit2Sharp/StatusOptions.cs
+++ b/LibGit2Sharp/StatusOptions.cs
@@ -35,6 +35,7 @@
         public StatusOptions()
         {
             DetectRenamesInIndex = true;
+            IncludeIgnored = true;
         }
 
         /// <summary>
@@ -84,5 +85,14 @@
         /// Unaltered meaning the file is identical in the working directory, the index and HEAD.
         /// </remarks>
         public bool IncludeUnaltered { get; set; }
+
+        /// <summary>
+        /// Include ignored files when scanning for status
+        /// </summary>
+        /// <remarks>
+        /// ignored meaning present in .gitignore. Defaults to true for back compat but may improve perf to not include if you have thousands of ignored files.
+        /// </remarks>
+        public bool IncludeIgnored { get; set; }
+
     }
 }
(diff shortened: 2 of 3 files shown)


paulgmiller:
Looks like you want commit coming in on the vnext branch. Don't see a way to edit target branch let me know if I just need to close this and start a new request with the same code.

ethomson:
Hi @paulgmiller !  Thanks for the submission.  I have a few comments that I will add inline, but we _do_ actually want this to target the **master** branch.

This is a relatively recent change, we used to use vNext.  Where did you see the guidance to open pull requests against vNext?  I need to update that.  😀 

Including ignored seems like an odd default.  I don't have the historical context to know why `IncludeIgnored` defaults to on here, but I don't love having different defaults from both libgit2 _and_ LibGit2Sharp.

It would be a bit of a breaking change to change the default, but properly documented, it might be the Right Thing.  @carlosmn ?

LibGit2Sharp/RepositoryStatus.cs
```diff
@@ -79,11 +79,15 @@ private static GitStatusOptions CreateStatusOptions(StatusOptions options)
                 Version = 1,
                 Show = (GitStatusShow)options.Show,
                 Flags =
-                    GitStatusOptionFlags.IncludeIgnored |
                     GitStatusOptionFlags.IncludeUntracked |
                     GitStatusOptionFlags.RecurseUntrackedDirs,
             };
 
+            if (options.IncludeIgnored )
```
ethomson: Minor nit: there's an additional space after `options.IncludeIgnored `...

LibGit2Sharp/StatusOptions.cs
```diff
@@ -84,5 +85,14 @@ public StatusOptions()
         /// Unaltered meaning the file is identical in the working directory, the index and HEAD.
         /// </remarks>
         public bool IncludeUnaltered { get; set; }
+
+        /// <summary>
+        /// Include ignored files when scanning for status
+        /// </summary>
+        /// <remarks>
+        /// ignored meaning present in .gitingore. Defaults to true for back compat but may improve perf to not include if you have thousands of ignored files.
```
ethomson: Typo in `.gitignore`.

paulgmiller:
Vnext branch is refrenced in https://github.com/libgit2/libgit2sharp/wiki under contributing items 1 and 7. 

Fixing up minor comments. Might be a good idea to make the default change separate so it can be reverted if need be. I'll run a git blame to try and see who introduced it but I obviously defer to you guys there.

paulgmiller:
Seems like it came in as the default in 974ad998 when you changed to track renames and moved away from git_status_foreach,

paulgmiller:
Anything else I can do here or are we waiting on @carlosmn's input on the default.

carlosmn:
Including ignored is probably to do with rename detection or somesuch. I don't like having that as default, but let's not change it for now.

ethomson:
Sounds fine.  I think that this is a crap default, and I think that it was my fault, and we should change it, but I think that it probably does make sense to change this in a separate PR.  (@paulgmiller are you interested in doing that?)

paulgmiller:
Maybe let me see how many UTs break. Unfortunately the product I noticed this in is considering moving to git.exe parsing rather than continuing to call libgit2sharp. Could have a longer conversation about why if you're interested

ethomson:
> Could have a longer conversation about why if you're interested

Feel free to drop by [our slack](http://slack.libgit2.org/) if you want to chat.

PR #1329, opened 2016-06-20
Fixed C Calling conversion.

zezba9000:
If you don't set "CallingConvention = CallingConvention.Cdecl" you will get runtime stack issues in .NET. The default is 'Winapi', which libgit2 isn't.

Also "UnmanagedType.Bool" is not the correct 'C' bool type "UnmanagedType.I1" is.

### DIFF
--- a/LibGit2Sharp/Core/NativeMethods.cs
+++ b/LibGit2Sharp/Core/NativeMethods.cs
@@ -150,7 +150,7 @@ namespace LibGit2Sharp.Core
             GitBuf buf,
             git_object* blob,
             [MarshalAs(UnmanagedType.CustomMarshaler, MarshalCookie = UniqueId.UniqueIdentifier, MarshalTypeRef = typeof(StrictFilePathMarshaler))] FilePath as_path,
-            [MarshalAs(UnmanagedType.Bool)] bool check_for_binary_data);
+            [MarshalAs(UnmanagedType.I1)] bool check_for_binary_data);
 
         [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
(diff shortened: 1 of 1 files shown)


ethomson:
libgit2 uses whatever calling conventions we ask it to.  We build _it_ with `-DSTDCALL=ON` for stdcall, not cdecl calling conventions.

We do not use cdecl because we make heavy use of callbacks, and while you can trivially annotate a pinvoke with the correct calling conventions, the callbacks are much more complicated to do so.

If you want to build your own libgit2 to use with LibGit2Sharp then either make sure you either use the build scripts or use the correct build flags.

If you think the bool marshaling is incorrect, please either update this PR with more explanation (links to docs that point out why the new type is correct) or open a new one and close this.

Thanks!

zezba9000:
I ran into this while building from source so I could link to the latest gitlib2 I built from source.

The reason I did this was **Git LFS** doesn't work with the Nuget package I pull in VS 2015. It does however work if I build gitlib2 and gitlib2sharp from source (and fix the calling conversion).

I see you're correct about the Bool as they are using an "int" for bools. Sorry about that.

zezba9000:
Ok works tnx. Guess the base git2 lib doesn't even support LFS without more work which doesn't seem to be documented... aaa o well

ethomson:
No, just like git-lfs is external to Git, it is also external to libgit2.

Like with Git, you will need to configure a filter to handle this.

ethomson:
If you have questions about how to set up a filter driver to support git-lfs please open a separate question.  But since this PR is resolved I'm going to close it.  :)

PR #1331, opened 2016-06-30
make excluding ignored files the default in status options (perf impr…

paulgmiller:
Switch the default as talked about. Only a few tests had to be changed.

### DIFF
--- a/LibGit2Sharp.Tests/CheckoutFixture.cs
+++ b/LibGit2Sharp.Tests/CheckoutFixture.cs
@@ -625,7 +625,7 @@ namespace LibGit2Sharp.Tests
                     "bin/some_ignored_file.txt",
                     "hello from this ignored file.");
 
-                Assert.Equal(1, repo.RetrieveStatus().Ignored.Count());
+                Assert.Equal(1, repo.RetrieveStatus(new StatusOptions { IncludeIgnored = true }).Ignored.Count());
 
                 Assert.Equal(FileStatus.Ignored, repo.RetrieveStatus(ignoredFilePath));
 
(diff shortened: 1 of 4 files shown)


LibGit2Sharp.Tests/CheckoutFixture.cs
```diff
@@ -652,7 +652,7 @@ public void ForceCheckoutRetainsIgnoredChanges()
                     "bin/some_ignored_file.txt",
                     "hello from this ignored file.");
 
-                Assert.Equal(1, repo.RetrieveStatus().Ignored.Count());
+                Assert.Equal(1, repo.RetrieveStatus(new StatusOptions { IncludeIgnored = true }).Ignored.Count());
```
ethomson: Can we add a similar test:  `Assert.Equal(0, repo.RetrieveStatus().Ignored.Count())` above this?

LibGit2Sharp.Tests/CheckoutFixture.cs
```diff
@@ -625,7 +625,7 @@ public void CheckoutRetainsIgnoredChanges()
...
```
ethomson: Can we add a similar test:  `Assert.Equal(0, repo.RetrieveStatus().Ignored.Count())` above this?

ethomson:
I'm 👍 on this.  @carlosmn ?

I think that we should responsibly deprecate this, though.  I think that we should hold on to this until after the next release, and immediately merge a note in the `CHANGELOG` and a deprecation warning to notify people of the change.

I'm not very familiar with all the c# attributes.  Is it appropriate to use `[Obsolete("...")]` here to get people to notice...?  It seems like it is _not_, but I'm not sure there's a superior alternative...

carlosmn:
I don't know that there is a good way to do this. Yes we should switch the default, but I don't think there's a way to warn about this without it looking like we're going to remove the whole thing. A line in CHANGELOG is probably the best sensible way we have.

I'm also not a fan of billing this a performance improvement since you can already do this and the change in behaviour should be made to match git's defaults.

PR #1339, opened 2016-07-13
Remove obsolete field.

txdv:
This field was removed from libgit2 a year and a half ago.

This is the libgit2 commit: https://github.com/libgit2/libgit2/commit/659cf2029f322ea876d663d85783b48945227e8f#diff-7c9f3aae8ea34d171fe8f44fbedde372L140

This is the libgit2sharp commit: https://github.com/libgit2/libgit2sharp/commit/4c5c088e8f9b1bb494ed1c088f5838f8d67031d9

### DIFF
--- a/LibGit2Sharp/Core/GitCloneOptions.cs
+++ b/LibGit2Sharp/Core/GitCloneOptions.cs
@@ -23,8 +23,6 @@ namespace LibGit2Sharp.Core
         public GitCloneLocal Local;
         public IntPtr CheckoutBranch;
 
-        public IntPtr signature; // Really a SignatureSafeHandle
-
         public IntPtr RepositoryCb;
         public IntPtr RepositoryCbPayload;
 


txdv:
This was exposed because I fiddled with the structure, making it bigger before this field.

carlosmn:
The calendar time doesn't really say much. Is this removed in the current version of libgit2 we target/bundle?

txdv:
Yes, it used to be [here](https://github.com/libgit2/libgit2/blob/master/include/git2/clone.h#L136-L143), but now it is gone.

txdv:
I linked master, but the commit at which NativeBinaries is target is missing it too: https://github.com/libgit2/libgit2/blob/381caf56908ad9cdb5959850a331d7cf2f952e63/include/git2/clone.h#L136-L143

txdv:
ping @carlosmn

PR #1346, opened 2016-08-03
Update libgit2 natives to 1.0.157

ethomson:
Use the new nuget package layout for .NET Core compatibility with
native binaries.

### DIFF
--- a/LibGit2Sharp/LibGit2Sharp.csproj
+++ b/LibGit2Sharp/LibGit2Sharp.csproj
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\packages\LibGit2Sharp.NativeBinaries.1.0.137\build\LibGit2Sharp.NativeBinaries.props" Condition="Exists('..\packages\LibGit2Sharp.NativeBinaries.1.0.137\build\LibGit2Sharp.NativeBinaries.props')" />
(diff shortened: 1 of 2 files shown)


ethomson:
Well, in what I'm sure will be SHOCKING NEWS, this has absolutely no effect for me when using a new-style (`project.json` based) project targeting .NET 4.5.1.

(Yay, nuget!)

@bording I think I've reached and surpassed both my limit of comprehension here and my patience with nuget.  Any chance you can provide me with some of your wisdom?

bording:
I'm getting ready to go out of town for the rest of the week, and won't have access to a computer, but I'll definitely take a look at it when I get back.

ethomson:
<3

ethomson:
Okay, it turns out that you can't use an RID of `win-x64`, for net45 you simply must use `win7-x64` and `win7-x86` regardless of what version of Windows you're actually on.

Don't try to understand, just let it wash over you.

In any case, we now produce a `LibGit2Sharp.NativeBinaries` package that has a layout of `runtimes/{rid}/natives`, which is compatible with a new-style MSBuild-less `project.json` world.  On old-school `.csproj` based builds, we still run the `.props` which will take the binaries for the appropriate platform out of the `NativeBinaries` location and place it in a folder `lib` beneath the build target.  (We do *not* emulate the `runtimes/` strategy when we do the copy, because that would leave us liable to regress #1170 , instead we stick with `lib`).

I tested this with old style (`.csproj`) projects on a couple of platforms (Windows and Mono) and with new style (`project.json`) project on Windows.

I'm going to merge this so that I can build a proper nuget prerelease package and upload it.

bording:
@ethomson Glad to see that you figured out the problem, despite the craziness around the RID usage. How did you manage to track it down to `win-x64` vs. `win7-x64` being the problem?

PR #1349, opened 2016-08-04
Update libgit2 natives to 1.0.160

ethomson:
Update libgit2 to native version 1.0.160 (https://github.com/libgit2/libgit2/commit/baa87dfc5033364ab5ac71099725f8f6c3e6361c).

### DIFF
--- a/LibGit2Sharp/Core/NativeMethods.cs
+++ b/LibGit2Sharp/Core/NativeMethods.cs
@@ -121,11 +121,6 @@ namespace LibGit2Sharp.Core
             git_repository* repo,
             [MarshalAs(UnmanagedType.CustomMarshaler, MarshalCookie = UniqueId.UniqueIdentifier, MarshalTypeRef = typeof(StrictFilePathMarshaler))] FilePath relative_path);
 
-        internal delegate int source_callback(
-            IntPtr content,
-            int max_length,
-            IntPtr data);
-
         [DllImport(libgit2)]
         internal static extern unsafe int git_blob_create_fromstream(
             out IntPtr stream,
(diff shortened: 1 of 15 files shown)


txdv:
How did I miss this...

Looks, good, are you planning on exposing the new APIs in a different commit?

LibGit2Sharp/Core/NativeMethods.cs
```diff
@@ -989,7 +976,7 @@ static NativeMethods()
 
         /* Push network progress notification function */
         internal delegate int git_push_transfer_progress(uint current, uint total, UIntPtr bytes, IntPtr payload);
-        internal delegate int git_packbuilder_progress(int stage, uint current, uint total, IntPtr payload);
+        internal delegate int git_packbuilder_progress(int stage, UInt32 current, UInt32 total, IntPtr payload);
```
txdv: Is this really needed? uint is always a synonym for System.UInt32

ethomson: Right, probably not.  Thanks!

ethomson:
> Looks, good, are you planning on exposing the new APIs in a different commit?

I don't have any plans to.  Is there something in particular that you're interested in?  Is that something that you'd want to tackle?

txdv:
Yeah, I would like to tackle it(exposnig all the new APIs) beginning with (Set|Get)UserAgent . Actually, I already started doing that, but I ran against https://github.com/libgit2/libgit2sharp/pull/1339 and kind of lost track...

carlosmn:
Unless we've deprecated something, in which case we should immediately provide the new way of doing it, I'd rather not conflate the binaries update with exposing new functionality.

LibGit2Sharp/packages.config
```diff
@@ -1,4 +1,4 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="LibGit2Sharp.NativeBinaries" version="1.0.157" targetFramework="net4" allowedVersions="[1.0.157]" />
+  <package id="LibGit2Sharp.NativeBinaries" version="1.0.160" targetFramework="net40" />
```
carlosmn: I think this should still have a `allowedVersions` to stop it from trying to "update" to an incompatible version of the binaries.

ethomson: Fixed.

LibGit2Sharp/Core/NativeMethods.cs
ethomson: Fixed.

PR #1352, opened 2016-08-15
Expose enable/disable cache setting

mitesch:
Expose the existing GIT_OPT_ENABLE_CACHING setting in libgit2.
See: https://github.com/libgit2/libgit2/blob/master/src/settings.c#L139


ethomson:
This seems reasonable, though the test isn't actually testing anything.  I'd be inclined to remove it, though there's also a nice opportunity to add the getter to libgit2.

Out of curiosity, what's the advantage you're getting by turning off caching?

mitesch:
I'll remove the test case. 

I'm seeing cache_evict_entries taking a while in performance traces. We're already caching odb backend entries longer term in a shared cache in managed code, so I hesitate to increase the cache size in libgit2. Our Repository and OdbBackend objects are short lived anyway, so I think we're just filling the cache and then throwing it away. 

My initial testing is showing 10-20+% improvement for merges on a very large repo, but I still have more testing to do.

ethomson:
Cleaned up the conflicts and merged this manually.  Thanks!

Issue #1356, opened 2016-08-25
LEAKS_IDENTIFYING and LEAKS_TRACKING are enabled in release build.


kik:
LEAKS_IDENTIFYING and LEAKS_TRACKING seem to be commented out in release build.

https://github.com/libgit2/libgit2sharp/blob/f916e79575bea0a99d3c67249090f51ff62d4e23/LibGit2Sharp/Core/Handles/Libgit2Object.cs#L8

ethomson:
Right you are!  Could you submit a quick pull request to comment those out?

PR #1357, opened 2016-08-25
comment out LEAKS_IDENTIFYER and LEAKS_TRACKING

kik:
these variables for test only. #1356


ethomson:
Ouch.  Thanks for this!

Issue #1362, opened 2016-09-06
Reference 'refs/notes/commits' not found when creating LibGit2Sharp.Repository object


splatteredbits:
Windows 7/PowerShell 5/libgit2sharp 0.22

When I try to create a `LibGit2Sharp.Repository` object, I get this exception [1]

Steps to reproduce:
1. Open PowerShell
2. Load the assembly: `add-type -Path .\LibGit2Sharp.dll`
3. Create a repository object: `New-object 'LibGit2Sharp.Repository' 'F:\Build\LibGit2Sharp'`

Note the exception.

[1]

```
Message        : Reference 'refs/notes/commits' not found
Data           : {libgit2.code, libgit2.category}
InnerException : 
TargetSite     : Void HandleError(Int32)
StackTrace     :    at LibGit2Sharp.Core.Ensure.HandleError(Int32 result) in c:\Git\libgit2sharp\LibGit2Sharp\Core\Ensure.cs:line 160
                    at LibGit2Sharp.Core.Proxy.git_foreach[T1,T2,TResult](Func`3 resultSelector, Func`2 iterator, GitErrorCode[] ignoredErrorCodes) in c:\Git\libgit2sharp\LibGit2Sharp\Core\Proxy.cs:line 3399
                    at LibGit2Sharp.NoteCollection.GetEnumerator() in c:\Git\libgit2sharp\LibGit2Sharp\NoteCollection.cs:line 41
HelpLink       : 
Source         : LibGit2Sharp
HResult        : -2146233088
```

splatteredbits:
I attached a debugger to see what is going on and it looks like the error happens when PowerShell tries to display the object. It enumerates (or tries to) the repository object and when there aren't any notes, we get the exception.

If I save the repository object to a variable, I don't get the error, e.g. `$repo = New-object 'LibGit2Sharp.Repository' 'F:\Build\LibGit2Sharp'`

What's the best way to gracefully handle when there aren't any notes?

carlosmn:
Fixed vi #1363

PR #1363, opened 2016-09-07
Fixed issue #1362: enumerating the notes collection for a repository …

splatteredbits:
…that has no notes throws an exception.

I'm not sure if this is the best solution, but the test reproduces the problem.

This fixes issue #1362 .


ethomson:
Thanks!

PR #1379, opened 2016-10-24
Fix MSBuild build with xbuild 12.0 engine (mono 4.2.1)

knocte:
If someone tried to build libgit2sharp with the xbuild 12.0
engine that comes with mono 4.2.1 [[1]](quite widespread given
that it's the version bundled with Ubuntu 16.04.x LTS), the
build would not work.

By supporting older versions of MS.NET/Mono, newer versions
of Mono wouldn't be supported (because newer versions of Mono
don't ship older .NET 4.0 profile).

This change similar to this recent commit in MonoAddins:
mono/mono-addins@a3efa4c

[1] $ mono --version
Mono JIT compiler version 4.2.1 (Debian 4.2.1.102+dfsg2-7ubuntu4)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
    TLS:           __thread
    SIGSEGV:       altstack
    Notifications: epoll
    Architecture:  amd64
    Disabled:      none
    Misc:          softdebug
    LLVM:          supported, not enabled.
    GC:            sgen

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@
 
 language: csharp
 mono:
- - 3.12.0
  - 4.2.3
 
 os:
--- a/LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
+++ b/LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
@@ -11,7 +11,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>LibGit2Sharp.Tests</RootNamespace>
     <AssemblyName>LibGit2Sharp.Tests</AssemblyName>
-    <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <TargetFrameworkProfile />
   </PropertyGroup>
--- a/LibGit2Sharp/LibGit2Sharp.csproj
+++ b/LibGit2Sharp/LibGit2Sharp.csproj
@@ -11,7 +11,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>LibGit2Sharp</RootNamespace>
     <AssemblyName>LibGit2Sharp</AssemblyName>
-    <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <TargetFrameworkProfile />
     <NuGetPackageImportStamp>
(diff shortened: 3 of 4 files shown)


ethomson:
I'm not sure that I understand, and you haven't really provided any details about this change.  We build _today_ with Mono 4.2.3 - this is, in fact, what we use for our CI builds in Travis.

Why can you not build on Mono 4.2.1?  Was Mono 4.2.1 damaged, regressed, or otherwise missing something that 4.2.3 fixed?  Does the default Debian install of Mono 4.2.1 not include the .NET 4.0 profile, but a full install does?

knocte:
> I'm not sure that I understand, and you haven't really provided any details about this change.
> We build today with Mono 4.2.3 - this is, in fact, what we use for our CI builds in Travis.

Oh, right, had missed that detail, that's why you guys couldn't see the problem.

> Why can you not build on Mono 4.2.1? Was Mono 4.2.1 damaged, regressed, or otherwise missing 
> something that 4.2.3 fixed? Does the default Debian install of Mono 4.2.1 not include the .NET 4.0
> profile, but a full install does?

Yes, that is exactly the issue. If you compile Mono from source, it will fill the older .NET profile directories with binary assemblies, and fill the .NET 4.5.x profile folder with the compiled assemblies of the build, as far as I understand. And as far as I know, that's the reason why, when packaged in Debian/Ubuntu, these profiles were not populated (as all packages from Debian need to come from open sources, no binary retrievals during the build are allowed).

ethomson:
Mono provides a .NET 4.0 (and 2.0) implementation that is open source, which is why this works on prior versions of Debian/Ubuntu.  

Whether these are installed by default, or on your system, I don't know.  https://launchpad.net/ubuntu/xenial/+source/mono lists all the available packages (including .NET 4.0).

Can you install libmono-corlib4.0-cli and the other .NET 4.0 packages?  Does that resolve your problem?

knocte:
> Can you install libmono-corlib4.0-cli and the other .NET 4.0 packages? Does that resolve your problem?

I guess you mean `libmono-corlib4.0-cil`. I tried and this is the result:

```
$ sudo apt install libmono-corlib4.0-cil
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Package libmono-corlib4.0-cil is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  mono-reference-assemblies-4.0 mono-devel

E: Package 'libmono-corlib4.0-cil' has no installation candidate
```

I'm now installing `mono-reference-assemblies-4.0`. If this really fixes the problem, it got me by surprise really! (In previous versions of ubuntu, installing mono was enough to have all profiles... (or at least monodevelop, which I got installed, and whose package should actually recommend this reference-assemblies package... I guess it's a packaging bug).

Issue #760, opened 2014-06-12
`PatchEntryChanges` test coverage


yorah:
#702 introduced new properties on `PatchEntryChanges`. It would be good to have unit tests covering them.

PR #1387, opened 2016-11-15
Test to show forward and backslashes in sub directories are handled c…

ornatwork:
…orrectly.

Ref
https://github.com/libgit2/libgit2sharp/issues/1075

### DIFF
--- a/LibGit2Sharp.Tests/FileHistoryFixture.cs
+++ b/LibGit2Sharp.Tests/FileHistoryFixture.cs
@@ -290,9 +290,38 @@ namespace LibGit2Sharp.Tests
             }
         }
 
-        #region Helpers
+        [Theory]
+        [InlineData("https://github.com/ornatwork/nugetpackages")]
+        public void HandlesSubDirectorySeperators(string url)
+        {
+          var scd = BuildSelfCleaningDirectory();
+          var clonedRepoPath = Repository.Clone(url, scd.DirectoryPath);
+          string subdir1 = "devbox";
+          string subdir2_double_backslash = subdir1 + Path.DirectorySeparatorChar + "content";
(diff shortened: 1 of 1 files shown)


ethomson:
Hi!  Thanks for the PR.  However, after looking more closely at #1075, I'm not certain that the issue is correct as reported.  (And if it is something that we want, then we would only want that behavior on Windows.)

This is because you can, in fact, have a file named `\` in a Git repository:

```
erg:Temp/testrepos/test% git init .
Reinitialized existing Git repository in /Users/ethomson/Temp/testrepos/test/.git/
zoidberg:Temp/testrepos/test% touch \\
zoidberg:Temp/testrepos/test% git add \\
zoidberg:Temp/testrepos/test% git status       
On branch master

Initial commit

Changes to be committed:
  (use "git rm --cached <file>..." to unstage)

	new file:   "\\"

zoidberg:Temp/testrepos/test% git ls-files --stage
100644 e69de29bb2d1d6434b8b29ae775ad8c2e48c5391 0	"\\"
```

So on at least non-Windows platforms where `\` is never a directory separator, we would want to ensure that `\` was not treated as a path separator.

My inclination is to say that we _never_ want `\` as a directory separator for Git repository paths, though.  Git paths are separated by `/`, regardless of what the platform uses for its filesystem.

ornatwork:
You know this better than myself, should I just wrap the test to run only on windows ( which I'm leaning towards )  or do you just want to close out https://github.com/libgit2/libgit2sharp/issues/1075  
or something else ?

ornatwork:
I was intending to create a new pull request with my checkin  https://github.com/libgit2/libgit2sharp/pull/1387/commits/8c9791b6a0df6224af4d8473b7903dd74466882a for https://github.com/libgit2/libgit2sharp/issues/760 but for whatever reason it landed on this PR

PR #1393, opened 2016-12-14
Add InludeUntracked and RecurseUntrackedDirs to StatusOptions

amaiorano:
These map directly to options of the same name on libgit2, and are useful when you don't want to bother retrieving them.

### DIFF
--- a/LibGit2Sharp/StatusOptions.cs
+++ b/LibGit2Sharp/StatusOptions.cs
@@ -36,6 +36,8 @@
         {
             DetectRenamesInIndex = true;
             IncludeIgnored = true;
+            IncludeUntracked = true;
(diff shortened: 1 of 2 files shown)


LibGit2Sharp/StatusOptions.cs
```diff
@@ -94,5 +101,9 @@ public StatusOptions()
         /// </remarks>
         public bool IncludeIgnored { get; set; }
 
+        /// <summary>
+        /// Include untracked files when scanning for status
+        /// </summary>
+        public bool InludeUntracked { get; set; }
```
Therzok: This should be Include, not Inlude

amaiorano:
Updated commit comment as well

PR #1397, opened 2016-12-23
Remove deprecated methods

carlosmn:
These were deprecated for the release of v0.23 so let's get rid of them for the
next release.


ethomson:
Via #1461

PR #1409, opened 2017-01-13
NativePath: use Assembly.CodeBase

### LINKED ISSUE #1389: Unable to load native DLL if path contains %2F
libgit2sharp is failing to load the native DLLs if the directory path contains `%2F`

How did I get in this state?  Our Jenkins build checks out GitHub repositories with the branch name in them.  In my case, the branch name was in the format `features/dev10` with a forward slash in the branch name.

That branch path in Windows ends up being `C:\jenkins\workspace\features%2Fdev10\`

This failure can be **easily reproduced** in libgit2sharp by setting the Output Path of `LibGit2Sharp.Tests` to `busted%2Fdirectory\`

I've tracked the problem down to `GlobalSettings.cs`, lines 22-31: 
```c#
static GlobalSettings()
{
    if (Platform.OperatingSystem == OperatingSystemType.Windows)
    {
        string managedPath = new Uri(Assembly.GetExecutingAssembly().EscapedCodeBase).LocalPath;
        nativeLibraryPath = Path.Combine(Path.Combine(Path.GetDirectoryName(managedPath), "lib"), "win32");
    }

    registeredFilters = new Dictionary<Filter, FilterRegistration>();
}
```

It looks like creating a `Uri` out of the current path turns the legit `%2F` in the directory name back into a forward slash.

After that, the DLL load mechanism fails to find the native `git2.dll` in `NativeMethods.cs` because it adds an invalid directory to the `PATH` environment variable

I hit this issue using [Cake](http://cakebuild.net/) and the [GitVersion](http://cakebuild.net/dsl/gitversion) tool which in turn uses libgit2sharp

In the meantime, I might be able to add the DLL directory to my PATH environment variable before calling GitVersion as a workaround.  Also not putting forward slashes in the git branch names will work around this issue with Jenkins putting the %2F in the directory name.

Stack trace below
```
14:35:43 Unhandled Exception: System.TypeInitializationException: The type initializer for 'LibGit2Sharp.Core.NativeMethods' threw an exception. ---> System.DllNotFoundException: Unable to load DLL 'git2-381caf5': The specified module could not be found. (Exception from HRESULT: 0x8007007E)
14:35:43    at LibGit2Sharp.Core.NativeMethods.git_libgit2_init()
14:35:43    at LibGit2Sharp.Core.NativeMethods.LibraryLifetimeObject..ctor()
14:35:43    at LibGit2Sharp.Core.NativeMethods..cctor()
14:35:43    --- End of inner exception stack trace ---
14:35:43    at LibGit2Sharp.Core.NativeMethods.RemoveHandle()
14:35:43    at LibGit2Sharp.Core.NativeMethods.LibraryLifetimeObject.Finalize()```

ethomson:
Use `Assembly.CodeBase` instead of `Assembly.EscapedCodeBase` since
the latter does not actually properly escape the path in a reversible
way.

eg, if a path contains a space, it will be encoded as `%20` in the
`EscapedCodeBase` however if a path contains a literal `%20` then it
will _also_ be encoded as `%20`.

Fixes #1389

### DIFF
--- a/LibGit2Sharp/GlobalSettings.cs
+++ b/LibGit2Sharp/GlobalSettings.cs
@@ -23,7 +23,26 @@ namespace LibGit2Sharp
         {
             if (Platform.OperatingSystem == OperatingSystemType.Windows)
             {
-                string managedPath = new Uri(Assembly.GetExecutingAssembly().EscapedCodeBase).LocalPath;
+                /* Assembly.CodeBase is not actually a correctly formatted
+                 * URI.  It's merely prefixed with `file:///` and has its
+                 * backslashes flipped.  This is superior to EscapedCodeBase,
+                 * which does not correctly escape things, and ambiguates a
+                 * space (%20) with a literal `%20` in the path.  Sigh.
+                 */
+                var managedPath = Assembly.GetExecutingAssembly().CodeBase;
+                if (managedPath == null)
+                {
+                    managedPath = Assembly.GetExecutingAssembly().Location;
+                }
+                else if (managedPath.StartsWith("file:///"))
+                {
+                    managedPath = managedPath.Substring(8).Replace('/', '\\');
+                }
+                else if (managedPath.StartsWith("file://"))
+                {
+                    managedPath = @"\\" + managedPath.Substring(7).Replace('/', '\\');
+                }
+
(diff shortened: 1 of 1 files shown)


sandorfr:
Is there any chance to see this released soon?

Issue #1415, opened 2017-02-03
Commands.Unstage(repo, filename) no longer works in v0.23.1


zezba9000:
Commands.Unstage(repo, filename) no longer works in **v0.23.1**.
It simply does nothing.
This used to work fine in the last version I think.

Commands.Stage(...) works fine.

ethomson:
Hmm - our integration tests don't seem to indicate a problem.  Can you provide a quick snippet and small sample repository that reproduces the problem that you're seeing?  Thanks!

zezba9000:
Here is the download link to the git project: https://www.dropbox.com/s/6aqd1sbphijb1ap/GitRepo.7z?dl=0

Ok so here is a super simple test console app on windows I did to verify the problem:

> using LibGit2Sharp;
> 
> namespace TestGitLib2Unstage
> {
> 	class Program
> 	{
> 		static void Main(string[] args)
> 		{
> 			using (var repo = new Repository(@"D:\TEMP\Git-It-GUI_Tester"))
> 			{
> 				Commands.Unstage(repo, "GitItGUI/StartPage.xaml.cs");
> 			}
> 		}
> 	}
> }

zezba9000:
I've also confirmed this on two different computers.

Just confirmed v0.22.1 is the last version to work using the legacy method:
```
using LibGit2Sharp;

namespace TestGitLib2Unstage
{
	class Program
	{
		static void Main(string[] args)
		{
			using (var repo = new Repository(@"D:\TEMP\GitRepo"))
			{
				//Commands.Unstage(repo, "Yahoo.txt");
				repo.Unstage("Yahoo.txt");
			}
		}
	}
}
```

zezba9000:
Anyone have any ideas on this? Huge blocker for me.
I can make builds etc if needed.

This is blocking my project: https://github.com/reignstudios/Git-It-GUI
Everything else is working.

norek:
I've the same problem, I describe it in Issue #1380. I think it's the same problem.

zezba9000:
I'm pretty sure this is a bug in libgit2 not libgit2sharp. I've looked at the source code and the results are coming directly from the native plugin.

I find it odd one of the most commonly used features has been broken for so long. This is why I wish there was one lib based git project supported by the git team and then a stdin/stdout tool chain was built on top of it. Compilers etc are moving this direction to as we end up having two different projects that do the same thing because the original wasn't designed right :(

ethomson:
[...]

PR #1420, opened 2017-02-12
Commands.Unstage: write the index

ethomson:
The `Unstage` command does not write the index; it should, to match the semantics of the prior `Repository.Unstage`, and to match `Commands.Stage`.


ethomson:
Fixes #1415

PR #1318, opened 2016-05-08
Target .NET Standard 1.3

AArnott:
Adds .NET Core support (by way of targeting .NET Standard 1.3). .NET Framework 4.0 is still a target, but we 
leverage VS2017 .NET SDK projects to multi-target both net40 and netstandard1.3 to maximize reach and enable libgit2sharp to work across platforms by way of .NET Core.

Closes #1293

PR #1431, opened 2017-04-01
Update test project dependencies

bording:
As part of preparing for #1318, this updates the unit test dependencies to the latest versions, which support .NET Standard targets.


bording:
Marking as [WIP] until I get the CI builds sorted out. Need to get them to use a newer version of NuGet.

LibGit2Sharp/LibGit2Sharp.csproj
```diff
@@ -11,7 +11,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>LibGit2Sharp</RootNamespace>
     <AssemblyName>LibGit2Sharp</AssemblyName>
-    <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
```
AArnott: We should be able to revert this. We can target net40 in the library and target net46 in the test project.

bording: You're right, that should be possible. Once I finally get the tests working on Travis, I'll go back and change that back.

bording:
And we're green!

I had to disable xUnit 2.x's test parallelization because it was causing file locking conflicts. It would be worth revisiting this in the future to speed up the test runs.

I also had to change up some tests that were using the `InlineData` attribute to pass in `IEnumerable<string>` values. On mono, it was failing with a error about not being able to convert from `System.Object[]` to `IEnumerable<string>`.

AArnott:
Thanks!

PR #1438, opened 2017-04-18
Native loading: simplify `git_libgit2_init` / `git_libgit2_shutdown`

ethomson:
Move the `git_libgit2_init` call _out_ of the object that contains the finalizer that does the corresponding `git_libgit2_shutdown`.  We do not want to create the object with that finalizer until we are sure that libgit2 has been loaded.

Otherwise, that finalizer will be called even when the constructor threw an exception (probably from `git_libgit2_init` not finding the native library).  In this case, when the finalizer is invoked, it will rethrow the `TypeInitializationException`, and exceptions from finalizers are not catchable.

By ensuring that the object with the finalizer is only created after `git_libgit2_init` succeeds, then we can ensure that any failures in `git_libgit2_init` remain catchable.

Now that our object's lifecycle ensures that we must have called `git_libgit2_init` in order for it to have been created, we can remove the unnecessary reference counting logic.  We previously refcounted the native methods, with each `SafeHandleBase` updating that reference count, [so that we did not call `git_libgit2_shutdown` before we were finished with the safe handle](https://github.com/libgit2/libgit2sharp/pull/361).  But this is obsolete, and this refcounting was unused outside of the load / unload scenario.  So we can simply do the unload in the finalizer, without trying to do any unnecessary work on the refcount.

### DIFF
--- a/LibGit2Sharp/Core/NativeMethods.cs
+++ b/LibGit2Sharp/Core/NativeMethods.cs
@@ -56,7 +56,7 @@ namespace LibGit2Sharp.Core
         }
 
         // Shutdown the native library in a finalizer.
-        private sealed class NativeShutdownObject : CriticalFinalizerObject
+        private sealed class NativeShutdownObject
         {
             ~NativeShutdownObject()
             {
--- a/LibGit2Sharp/Core/NativeMethods.cs
+++ b/LibGit2Sharp/Core/NativeMethods.cs
@@ -17,58 +17,28 @@ namespace LibGit2Sharp.Core
     {
         public const uint GIT_PATH_MAX = 4096;
         private const string libgit2 = NativeDllName.Name;
-        // This is here to keep the pointer alive
-#pragma warning disable 0414
-        private static readonly LibraryLifetimeObject lifetimeObject;
-#pragma warning restore 0414
-        private static int handlesCount;
 
-        /// <summary>
-        /// Internal hack to ensure that the call to git_threads_shutdown is called after all handle finalizers
-        /// have run to completion ensuring that no dangling git-related finalizer runs after git_threads_shutdown.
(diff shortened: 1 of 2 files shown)


LibGit2Sharp/Core/NativeMethods.cs
```diff
@@ -14,71 +14,50 @@ internal static class NativeMethods
...
-        private static int handlesCount;
-
-        /// <summary>
-        /// Internal hack to ensure that the call to git_threads_shutdown is called after all handle finalizers
-        /// have run to completion ensuring that no dangling git-related finalizer runs after git_threads_shutdown.
-        /// There should never be more than one instance of this object per AppDomain.
-        /// </summary>
-        private sealed class LibraryLifetimeObject : CriticalFinalizerObject
+        private static readonly NativeShutdownObject shutdownObject;
+
```
AArnott: You no longer restore this warning. Was that intentional?

ethomson: It was not - thank you for catching this.

AArnott:
BTW this creates a merge conflict with #1318 so once you merge I'll do the best I can to resolve the conflict and update that PR, but please consider reviewing the merge commit to make sure I kept what you are trying to do here.

ethomson:
Yeah - I feel like everything's going to create a merge conflict with #1318.  😛 

I'm perfectly happy to wait on this until #1318 lands and deal with the merge conflicts myself - and I feel like we're very, very close to seeing #1318 land.  Since you were the motivation behind this PR, I'm going to say that it's really up to you.

AArnott:
I'm willing to go either way, but I guess it's easier for me if you hold off on merging this till #1318 is done. 

Thanks.

ethomson:
Will do - thanks @AArnott !

Issue #1451, opened 2017-05-18
Crash when using GlobalSettings.Version.ToString() in 0.23.1


mminns:
In 0.23.1 the generated libgit2sharp_hash.txt file contains 'HEAD' rather than a commit hash.

If you call LibGit2.GlobalSettings.Version.ToString() this causes a crash as it tries to extract 7 characters from a 4 character string

see https://github.com/libgit2/libgit2sharp/blob/master/LibGit2Sharp/Version.cs#L65 and https://github.com/libgit2/libgit2sharp/blob/master/LibGit2Sharp/Version.cs#L72

ethomson:
Thanks for reporting this and for the fix!

PR #1452, opened 2017-05-20
Allow Version class to deal with HEAD as a version hash

mminns:
The Version code expects a commit Hash to be used to specifiy the version of code in use in the library. However this can also be HEAD. If it is HEAD the current code will crash as it tries to extract 7 characters out of 4.

There are no new unit tests, the existing ones pass, since it would require a mechanism to re-write or mock the _hash.txt files during the tests, which seemed beyond the scope of this change.

see https://github.com/libgit2/libgit2sharp/issues/1451


ethomson:
Manually merged - thanks for the fix!

PR #1456, opened 2017-06-13
Update libgit2 to 2ca088b

ethomson:


LibGit2Sharp/packages.config
```diff
@@ -1,4 +1,4 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="LibGit2Sharp.NativeBinaries" version="1.0.165" targetFramework="net40" allowedVersions="[1.0.165]" />
-</packages>
+  <package id="LibGit2Sharp.NativeBinaries" version="1.0.183" targetFramework="net40" />
```
bording: You're missing the `allowedVersions` attribute here.

ethomson:
Hmm.  Looks like the Unix platforms aren't loading their native libraries...  they're in the nuget and, from the logs, it looks like they're getting deployed correctly.

@bording any ideas why this might happen?

bording:
Hmm, Travis is currently configured to use the latest version of mono, which is now 5.0. I know it was working fine with 4.8, so maybe try modifying the Travis config to run against 4.8 and see if that's what has changed?

ethomson:
Huh.  Good call!

bording:
So now the question is what changed in 5.0 to break this? I'll play around with it a bit to try and figure out what the problem is.

PR #1469, opened 2017-07-08
Fix CertificateSsh constructor

### LINKED ISSUE #1383: CertificateSsh constructor looks wrong
While reading through the LibGit2Sharp code, I noticed something that looks wrong to me. Here's the code as it is currently ([lines 48-55 of `CertificateSsh.cs`](https://github.com/libgit2/libgit2sharp/blob/38f046d83da3fb98b2d57fca1d428363dc31a960/LibGit2Sharp/CertificateSsh.cs#L48-L55)):

``` csharp
            HashMD5 = new byte[16];
            fixed (byte* p = &HashMD5[0])
            {
                for (var i = 0; i < HashMD5.Length; i++)
                {
                    HashMD5[i] = p[i];
                }
            }
```

That looks like it's _intended_ to copy the MD5 hash from the `git_certificate_ssh` struct passed in as the `cert` parameter. But what it's _actually_ doing is to allocate a new 16-byte array, and copy it onto itself, basically performing a no-op.

I think that `fixed (byte* p = &HashMD5[0])`  should be changed to `fixed (byte* p = &cert->HashMD5[0])`. And likewise with [the next `fixed` block](https://github.com/libgit2/libgit2sharp/blob/38f046d83da3fb98b2d57fca1d428363dc31a960/LibGit2Sharp/CertificateSsh.cs#L57-L64): `fixed (byte* p = &HashSHA1[0])` should be changed to `fixed (byte* p = &cert->HashSHA1[0])`.

rmunn:
Should fix #1383.

If this looks too simple, see https://stackoverflow.com/q/44870128/ where I asked about this code to make sure that I wasn't missing something. Three people (the one who commented and the two who upvoted it) agreed that this simple-looking code is fine: there was no need for pointer arithmetic in this constructor. (And there is probably no need for pointer arithmetic in the `ToPointer()` method either, but one fix at a time).

### DIFF
--- a/LibGit2Sharp/CertificateSsh.cs
+++ b/LibGit2Sharp/CertificateSsh.cs
@@ -35,10 +35,6 @@ namespace LibGit2Sharp
         /// </summary>
         public readonly bool HasSHA1;
 
-        /// <summary>
-        /// True if we have the SHA1 hostkey hash from the server
-        /// </summary>public readonly bool HasSHA1;
-
         internal unsafe CertificateSsh(git_certificate_ssh* cert)
         {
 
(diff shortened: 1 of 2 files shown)


ethomson:
Cool, thanks @rmunn for the fix, and @bording for the review!

PR #1474, opened 2017-07-20
Add preview information to version number

ethomson:
Change the version numbers to 0.25.0-preview-nnnn, which is basically in line with our old preview number strategy.

### DIFF
--- a/LibGit2Sharp/LibGit2Sharp.csproj
+++ b/LibGit2Sharp/LibGit2Sharp.csproj
@@ -32,8 +32,8 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="LibGit2Sharp.NativeBinaries" Version="[1.0.185]" PrivateAssets="contentFiles" />
-    <PackageReference Include="SourceLink.Create.GitHub" Version="2.0.2" PrivateAssets="all" />
-    <DotNetCliToolReference Include="dotnet-sourcelink-git" Version="2.0.2" />
+    <PackageReference Include="SourceLink.Create.GitHub" Version="2.1.2" PrivateAssets="all" />
+    <DotNetCliToolReference Include="dotnet-sourcelink-git" Version="2.1.2" />
 
     <PackageReference Include="CodeGeneration.Roslyn.BuildTime" Version="$(CodeGenerationRoslynVersion)" PrivateAssets="all" />
     <DotNetCliToolReference Include="dotnet-codegen" Version="$(CodeGenerationRoslynVersion)" />
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -3,10 +3,10 @@
     <OutputPath>$(MSBuildThisFileDirectory)bin\$(MSBuildProjectName)\$(Configuration)\</OutputPath>
     <BaseIntermediateOutputPath>$(MSBuildThisFileDirectory)obj\$(MSBuildProjectName)\</BaseIntermediateOutputPath>
 
-    <CodeGenerationRoslynVersion>0.4.9</CodeGenerationRoslynVersion>
+    <CodeGenerationRoslynVersion>0.4.11</CodeGenerationRoslynVersion>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Nerdbank.GitVersioning" Version="1.6.25" PrivateAssets="all" />
+    <PackageReference Include="Nerdbank.GitVersioning" Version="2.0.37-beta" PrivateAssets="all" />
   </ItemGroup>
 </Project>
--- a/LibGit2Sharp.Tests/GlobalSettingsFixture.cs
+++ b/LibGit2Sharp.Tests/GlobalSettingsFixture.cs
@@ -30,7 +30,7 @@ namespace LibGit2Sharp.Tests
             //      git2hash: '06d772d' LibGit2 library hash.
             //      arch: 'x86' or 'x64' LibGit2 target.
             //      git2Features: 'Threads, Ssh' LibGit2 features compiled with.
-            string regex = @"^(?<version>\d+\.\d+\.\d+(-\w+)?\+(g(?<git2SharpHash>[a-f0-9]{10})\.)?LibGit2-[a-f0-9]{7}) \((?<arch>\w+) - (?<git2Features>(?:\w*(?:, )*\w+)*)\)$";
+            string regex = @"^(?<version>\d+\.\d+\.\d+(-[\w\-\.]+)?\+(g(?<git2SharpHash>[a-f0-9]{10})\.)?LibGit2-[a-f0-9]{7}) \((?<arch>\w+) - (?<git2Features>(?:\w*(?:, )*\w+)*)\)$";
 
             Assert.NotNull(versionInfo);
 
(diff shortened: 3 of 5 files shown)


bording:
Given that we aren't using a more complicated branching strategy here, like git flow, this seems like an OK change to make.

However, if we do actually want to release a prerelease of some sort (for example to get a .NET Core compatible version on NuGet), how would we do that?

ethomson:
Right - my goal here was explicitly to get a .NET Core compatible nuget package on nuget.org.  This didn't really solve that problem, you're quite right.  Now we have an 0.24.2-gwhatever version.

Bleh.

@AArnott what's up with this version number strategy you have here?  I want to make 0.25 prereleases, and have a version strategy for 0.24 hotfixes.  How should this work?

AArnott:
The basic philosophy behind NB.GV is that every git commit should produce a build with a unique version stamp. That way, any binary can be traced back to its exact source code trivially. In addition when it comes to NuGet package versions, this ensures that NuGet caches don't start giving ghostly problems where you have several commits in your repo that all build the same package version, only one of which can exist in your nuget cache at once and thus you think you're getting one build installed for testing but you actually get another instead and it takes you hours of debugging before you realize it.

So now lets analyze each concern you brought up:

> Now we have an 0.24.2-gwhatever version.

The `-gc0ffee` suffix only appears on non-shipping branches. Any build from [one of master or a vX.Y branch](https://github.com/libgit2/libgit2sharp/blob/master/version.json#L5-L6) will *not* have the `-g*` suffix appended. The suffix is basically there so that any topic branches get a guaranteed unique version number

> But the new system is incrementing the teeny version, so now we're at "0.25.14" even though we've never released an "0.25"

That's the default behavior in order to ensure each build has a unique version. Yes, that means you'll ship 0.25.14-beta and later ship 0.25.40-beta, and finally when you ship an 'RTW' release it will be 0.25.55. And servicing it will make it 0.25.56, 0.25.57, and so on.
In my experience, this never produces customer confusion or complaints, and it certainly helps understand what version folks are really using. So my recommendation would be to just go with it. I would strongly recommend *against* going back to 0.24 versions because you're not building 0.24 any more by intention -- you're actually working on 0.25 prereleases so your version should reflect that.

If you really don't like this incrementing of the 3rd integer though, NB.GV recently advanced to offer you another option as documented [here](https://github.com/AArnott/Nerdbank.GitVersioning/blob/master/doc/versionJson.md#file-format). Specifically, you can take control of all 3 version integers, and move the git height to the prerelease tag, so it looks like `1.2.3-beta.{height}` and eventually you can remove the prerelease tag and just ship `1.2.3` if that's what you really want. Not specifying `{height}` anywhere greatly erodes the value of never having two packages with the same version though, so I recommend you go with a strategy that retains height somewhere.

bording:
@AArnott You mentioned 0.25.14-beta and 0.25.40-beta, but how do we actually get that -beta tag on a release?

AArnott:
Just add it to the `version.json` file's `version` field:

```json
{
  "version": "0.25-beta"
}
```

bording:
And that would know to insert the patch version in the correct place?

bording:
@ethomson I like the idea of including the height in the prerelease tag, seems to be best all around solution, because you get full control over all 3 parts of the version number, and you get the auto-incrementing part as well.

ethomson:
Agree that we don't want to revert to 0.24.0 - as you both pointed out, that's definitely not solving the problem.  We want to keep 0.24.teeny for actual hotfixes to the 0.24 release.

I do prefer to be able to release 0.25 as 0.25.0, mostly just because I'm old school.  Thanks for the insight here @AArnott .

AArnott:
@ethomson If you intend to ship 0.25.0, then you shouldn't build 0.25.15-preview leading up to that. So your version.json change should set version to `0.25.0-preview.{height}`.

AArnott:
For example, your PR is building `0.25.30-preview+g9db0678caf` whereas you probably want to see `0.25.0-preview.30+g9db0678caf`

ethomson:
Awesome, thanks!

AArnott:
Oh, from the build break I see we're still using NB.GV 1.6.25. The ability to move the `{height}` to other locations is new in the 2.0 versions. So you should upgrade the NB.GV package to `2.0.37-beta`.

CodeGenerationAttributes/CodeGenerationAttributes.csproj
```diff
@@ -7,6 +7,10 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="CodeGeneration.Roslyn.Attributes" Version="$(CodeGenerationRoslynVersion)" />
+    <PackageReference Include="CodeGeneration.Roslyn.Attributes" Version="0.4.11" />
```
bording: Instead of modifying these values here and other places, you should update the [Directory.Build.props](https://github.com/libgit2/libgit2sharp/blob/master/Directory.Build.props) file instead.

CodeGenerationAttributes/CodeGenerationAttributes.csproj
```diff
@@ -7,6 +7,10 @@
...
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Update="Nerdbank.GitVersioning" Version="2.0.37-beta" />
```
bording: This is already referenced in the Directory.Build.props file, so you should update that one reference and remove the ones you added to the individual project files.

CodeGenerationAttributes/CodeGenerationAttributes.csproj
ethomson: Oh huh.  Thanks for catching this, I just ran update via VS.

bording: Yeah, the tooling inside of VS doesn't really understand the Directory.Build.props/targets files yet. 😖

ethomson: Good to know.  I fixed it up manually, should be good now.

LibGit2Sharp/LibGit2Sharp.csproj
```diff
@@ -32,7 +32,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="LibGit2Sharp.NativeBinaries" Version="[1.0.185]" PrivateAssets="contentFiles" />
-    <PackageReference Include="SourceLink.Create.GitHub" Version="2.0.2" PrivateAssets="all" />
+    <PackageReference Include="SourceLink.Create.GitHub" Version="2.1.2" PrivateAssets="all" />
```
bording: Sorry, didn't notice this before, but the `dotnet-sourcelink-git` tool reference below should be bumped to 2.1.2 as well to match this change.

bording:
@ethomson I went ahead and fixed the `dotnet-sourcelink-git` version and pushed up a rebased branch.

bording:
I've restarted the mac build a few times, but it seems Travis is having connectivity issues again on their macs...

ethomson:
Awesome, thanks @bording

ethomson:
Hrmf.  Well, I don't know why the Mac build isn't working, but it's unrelated to this, so I'm going to merge it.  Thanks again for your help y'all.

PR #1501, opened 2017-10-26
Add SubmoduleCollection.Add

JasonGoemaat:
Added `SubmoduleCollection.Add(url, path, useGitLink)`.  

Looking for some input here if this can be done a better way.  The docs say call `git_submodule_add_setup()`, then clone the repo or whatever, then call `git_submodule_add_finalize()` to add the submodule and '.gitmodules' to the index.

I wasn't sure how to create a `Submodule` without a name, or is that just the path?  It seems like cloning the repository into the desired directory first and then calling both setup and finalize methods works fine and is how it is actually done when creating a new submodule, isn't it?


jminer:
I need this functionality too. What needs to be done to get this added? I can probably work on it some if needed. I saw in a past PR a review comment that this code should be in `Commands.SubmoduleAdd`. Is that still true? Is there anything else?

PR #1502, opened 2017-11-08
README: point coveralls to master branch

ethomson:
(We haven't done vNext for years, ugh.)


ethomson:
One could make a reasonable argument that coveralls (or code coverage in general) is pointless at best or misleading at worst, but that's a different PR.

PR #1505, opened 2017-11-10
Expose enable/disable ofs delta and strict object creation settings

jianges:
Expose the existing GIT_OPT_ENABLE_OFS_DELTA and GIT_OPT_ENABLE_STRICT_OBJECT_CREATION setting in libgit2.
See https://github.com/libgit2/libgit2/blob/master/src/settings.c#L225 and https://github.com/libgit2/libgit2/blob/master/src/settings.c#L193

### DIFF
--- a/LibGit2Sharp/Core/Proxy.cs
+++ b/LibGit2Sharp/Core/Proxy.cs
@@ -3429,7 +3429,29 @@ namespace LibGit2Sharp.Core
             Ensure.ZeroResult(res);
         }
 
-#endregion
+        /// <summary>
+        /// Enable or disable the ofs_delta capabilty
+        /// </summary>
+        /// <param name="enabled">true to enable the ofs_delta capabilty, false otherwise</param>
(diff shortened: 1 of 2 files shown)


LibGit2Sharp/Core/Proxy.cs
```diff
@@ -3429,7 +3429,29 @@ public static void git_libgit2_opts_set_enable_caching(bool enabled)
...
 
-#endregion
+        /// <summary>
+        /// Enable or disable the ofs_delta capabilty
+        /// </summary>
+        /// <param name="enabled">true to enable the ofs_delta capabilty, false otherwise</param>
+        public static void git_libgit2_opts_set_enable_ofsdelta(bool enabled)
+        {
+            // libgit2 expects non-zero value for true
+            var res = NativeMethods.git_libgit2_opts((int)LibGitOption.EnableOfsDelta, (uint)(enabled ? 1 : 0));
```
ethomson: I think this should be `LibGit2Option`.

ethomson:
Thanks @jianges!

PR #1509, opened 2017-11-17
Update Dependencies

ethomson:
The current version of xunit's Visual Studio integration has been _incredibly_ flaky on my machines.  So far, the Xunit 2.3 betas have been much more stable.  Now that there's a proper Xunit version, let's upgrade our dependencies to use it.

### DIFF
--- a/LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
+++ b/LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
@@ -18,12 +18,16 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Moq" Version="4.7.99" />
-    <PackageReference Include="xunit" Version="2.2.0" />
+    <PackageReference Include="Moq" Version="4.7.145" />
+    <PackageReference Include="xunit" Version="2.3.1" />
     <PackageReference Include="xunit.skippablefact" Version="1.3.3" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.2.0" />
-    <PackageReference Include="xunit.runner.console" Version="2.2.0" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.3.1" />
+    <PackageReference Include="xunit.runner.console" Version="2.3.1" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.5.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Update="Nerdbank.GitVersioning" Version="2.0.41" />
   </ItemGroup>
 
 </Project>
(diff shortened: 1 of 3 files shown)


LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
```diff
@@ -18,12 +18,16 @@
...
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.2.0" />
-    <PackageReference Include="xunit.runner.console" Version="2.2.0" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.3.1" />
+    <PackageReference Include="xunit.runner.console" Version="2.3.1" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.5.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Update="Nerdbank.GitVersioning" Version="2.0.41" />
```
bording: This is currently in the Directory.Build.props file, so we should either update that one entry, or remove it if we want it in the project files after all.

LibGit2Sharp/LibGit2Sharp.csproj
```diff
@@ -34,10 +34,14 @@
 
   <ItemGroup>
     <PackageReference Include="LibGit2Sharp.NativeBinaries" Version="[1.0.185]" PrivateAssets="contentFiles" />
-    <PackageReference Include="SourceLink.Create.GitHub" Version="2.2.0" PrivateAssets="all" />
+    <PackageReference Include="SourceLink.Create.GitHub" Version="2.5.0" PrivateAssets="all" />
     <DotNetCliToolReference Include="dotnet-sourcelink-git" Version="2.2.0" />
```
bording: The DotNetCliToolReference also needs to be bumped to 2.5.0.

LibGit2Sharp/LibGit2Sharp.csproj
```diff
@@ -34,10 +34,14 @@
...
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Update="Nerdbank.GitVersioning" Version="2.0.41" />
```
bording: Same as above. This is already in the Directory.Build.props file.

I'd say there's an argument that we don't need it there, since this is the only project that actually cares about the version, and having it in the props file means it applies to the test project as well.

If we do keep it here instead, the extra ItemGroup should be removed and it should go up with the other packages above. It also needs `PrivateAssets="all"` added.

LibGit2Sharp/LibGit2Sharp.csproj
```diff
@@ -34,10 +34,14 @@
 
   <ItemGroup>
     <PackageReference Include="LibGit2Sharp.NativeBinaries" Version="[1.0.185]" PrivateAssets="contentFiles" />
```
bording: We should go ahead and build a new native binaries package that we can update to as well. 1.0.185 doesn't have all of the changes I made around support netstandard.

bording:
Superseded by #1511

PR #1510, opened 2017-11-17
WIP: Rebase Refactoring

ethomson:
Goals:
1. Move `Rebase` into the commands that we've started.
2. Provide the ability to do in-memory / bare rebases.

### DIFF
--- a/LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
+++ b/LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
@@ -18,12 +18,16 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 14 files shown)


BKSpurgeon:
Hi there
first of all thank you for your contribution to the code base

I would like to know re: this refactoring - what is the reason/purpose for it?

rgds

Ben

ethomson:
Ideally users would be able to do a rebase completely in memory, without touching the working directory, and with custom ODB, etc backends. This is well supported by libgit2 but LibGit2Sharp makes various assumptions and these configurations are difficult or impossible.

BKSpurgeon:
ok chrs. appreciate the comment.

PR #1511, opened 2017-11-19
Remove the net40 target and other misc cleanup

bording:
This PR includes the following changes;

The net40 target has been removed from LibGit2Sharp, so we are now targeting .NET Standard 2.0 only. This means that our minimum supported version of the .NET Framework is now 4.6.1.

The test project is still targeting both net461 and netcoreapp2.0 so that we can test on both platforms.

Dependencies have been updated (this supersedes #1509), and I've moved the Nerdbank.GitVersioning reference into the main project file to make it easier to update properly with the NuGet tooling.

With the new version of xunit, there were a ton of warnings in the tests, so I cleaned up all of those as well.

As of the 2.0.2 .NET Core SDK that is installed with VS 2017 15.4, there's a simpler way to include PDBs in the package, so I've switched to that.

I've also updated to the version of the native binaries package that includes my changes from https://github.com/libgit2/libgit2sharp.nativebinaries/pull/56.

The build/CI scripts still need some cleaning up, but I'll handle that in another PR.


bording:
@ethomson Once this is merged, I think we should put out another 0.25.0 preview package.

bording:
One more change to make the GenerateAssemblyCommitIdsCs target handle when the project is being built from outside of a git repo, which should address #1512.

ethomson:
Appreciate you setting it up so that you could build from a release tarball/zipball.

ethomson:
LGTM.  Thanks!

PR #1517, opened 2017-11-23
Remove two Console.WriteLine calls from the test fixture

carlosmn:


bording:
@carlosmn Were these causing a problem, or are they just unnecessary?

carlosmn:
They're causing useless, context-less output when running the tests.

ethomson:
Lol wut

Issue #1515, opened 2017-11-22
CreateBlob fails on files greater than 5 gigs


halfpeaw:
Verified using the 64bit version of LibGit2sharp

Created a file stream that refers to a large 5 gig pst file.
`Stream readStream = new FileStream(filePath, FileMode.Open);`
`Blob newBlob = repo.ObjectDatabase.CreateBlob(readStream, readStream.Length);`

Got this error 
LibGit2Sharp.LibGit2SharpException: failed to finish zlib inflation; stream aborted prematurely

ethomson:
In a cruel twist, this isn't failing to create a blob in the repository, it's failing to load the blob _after_ it's been created.  Hence the `failed to finish zlib `**`inflation`** error (not deflate).

1. This API is not ideal for you, unless you really do need the resultant `Blob` object (with all 5 GB loaded into memory).  You want an API that writes to the object database and just gives you back the oid instead of trying to reread it.
2. Unfortunately, the only API that does _that_ expects you to hand it the entire `byte[]` to write.  Ugh.
3. I don't know why this particular call is failing (why we can't `Blob.Lookup` a 5 GB object) but unless that's actually important to you, I'm going to worry about that part later.

ethomson:
I created #1518 to provide a mechanism to write a stream to the object database.  I hope that this resolves your problems, @halfpeaw.  However, the `Blob.Lookup` issue remains.

carlosmn:
There is a rather persistent idea across the bindings that objects are lightweight, which makes some things a bit complicated.

IIRC we used to just keep the id of the blob in the `Blob` object until you tried to load it. This caused different issues so we went with actually loading stuff instead, and I suspect a lot of these uses are a left-over from that time (for compat with types and all that).

ethomson:
I've created #1518 and #1519 which should help you here, @halfpeaw.  However, I'm still seeing some issues with large files going into the ODB, so I'm still digging.

ethomson:
So it seems that Git for Windows _also_ has an issue decompressing one of the large files that I created ((the file was basically uncompressible, so the zip file itself was >5GB).

So as far as a situation report goes:  it looks like LibGit2Sharp should be able to add very large files to the object database properly, and add them to a tree.  I'll dig in to why we fail reading next, but I think that this should unblock your use case @halfpeaw and I'll create a prerelease nuget package.

PR #1518, opened 2017-11-24
Introduce ObjectDatabase.Write(Stream...)

ethomson:
Introduce ObjectDatabase.Write(Stream...) to write directly to the object database (without creating the subsequent `Blob`) and without having to load the entire blob into memory.  See #1515

### DIFF
--- a/LibGit2Sharp/ObjectDatabase.cs
+++ b/LibGit2Sharp/ObjectDatabase.cs
@@ -178,7 +178,7 @@ namespace LibGit2Sharp
         }
 
         /// <summary>
-        /// Write an object to the object database
+        /// Writes an object to the object database.
         /// </summary>
         /// <param name="data">The contents of the object</param>
         /// <typeparam name="T">The type of object to write</typeparam>
--- a/LibGit2Sharp/ObjectDatabase.cs
+++ b/LibGit2Sharp/ObjectDatabase.cs
@@ -188,6 +188,45 @@ namespace LibGit2Sharp
         }
 
         /// <summary>
+        /// Writes an object to the object database.
+        /// </summary>
+        /// <param name="stream">The contents of the object</param>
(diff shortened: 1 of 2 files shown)


LibGit2Sharp/ObjectDatabase.cs
```diff
@@ -187,6 +187,45 @@ public int Provider(IntPtr content, int max_length, IntPtr data)
             return Proxy.git_odb_write(handle, data, GitObject.TypeToKindMap[typeof(T)]);
         }
 
+        /// <summary>
+        /// Write an object to the object database
```
bording: Nitpick, but this should be "Writes" instead of "Write" to be consistent with other comments.

Also, all of these should have a period at the end.

carlosmn:
If we're fixing the API, we should make the `Write` call return something that implements (or inherits from, in this language) `Stream`.

This API comes from when we had the scatter-gather pull mechanism in libgit2 for writing to the objectdb. We kept the API for compat when we let you push data into libgit2, but any new code should ideally not use that.

ethomson:
Why would we return a `Stream`?

ethomson:
Right, I understand what you're saying now, @carlosmn: instead of taking a `Stream` that `CanRead` and pulling the contents out of it and into the ODB, we could return a `Stream` that `CanWrite` and people can write to it at their leisure and we will put the contents into the ODB.

I think that moving towards that API might make sense.  At the moment, though, this API matches our existing APIs and unblocks some scenarios.

LibGit2Sharp/ObjectDatabase.cs
ethomson: To be fair, it's not _that_ consistent, seeing as I copied and pasted the documentation from the other `Write` method above it.  😝 

Updated them both, in any case.

PR #1519, opened 2017-11-25
TreeDefinition: allow `Add` a blob by ID

ethomson:
Allow consumers to add a blob to a `TreeDefinition` by specifying only the object ID and mode.  This lets users build tree entries without having to create a `Blob` object (and thus load the object itself) which is beneficial for users building trees with large objects.

This is only useful for Blobs, since they do not need to be fetched from the object database to be realized within the tree builder.  Users are not able to add Trees by ID, since we would need to load them (and cannot, since we are not guaranteed to be instantiated within the context of a repository).

See #1515

### DIFF
--- a/LibGit2Sharp.Tests/TreeDefinitionFixture.cs
+++ b/LibGit2Sharp.Tests/TreeDefinitionFixture.cs
@@ -266,6 +266,49 @@ namespace LibGit2Sharp.Tests
             }
         }
 
+        [Theory]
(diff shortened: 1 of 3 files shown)


ethomson:
I don't love the asymmetry here (ie, that you cannot add a `Tree` by its ID) but I would like to do some bigger refactoring on this class later.  Until then, I'm going to :shipit: this.

PR #1534, opened 2018-02-12
Add MergeCommitsIntoIndex to ObjectDatabase

tyrielv:
Adds a new method MergeCommitsIntoIndex to ObjectDatabase to make the intermediate result of MergeCommits available (so that conflicts can be resolved directly against the produced index).

Also adds WriteToTree to Index so that the produced index can be converted to a tree just like MergeCommits currently does.

### DIFF
--- a/LibGit2Sharp/ObjectDatabase.cs
+++ b/LibGit2Sharp/ObjectDatabase.cs
@@ -856,8 +856,9 @@ namespace LibGit2Sharp
         /// <param name="ours">The first tree</param>
         /// <param name="theirs">The second tree</param>
         /// <param name="options">The <see cref="MergeTreeOptions"/> controlling the merge</param>
-        /// <returns>The <see cref="Index"/> containing the merged trees and any conflicts, or null if the merge stopped early due to conflicts</returns>
-        public virtual Index MergeCommitsIntoIndex(Commit ours, Commit theirs, MergeTreeOptions options)
+        /// <returns>The <see cref="TransientIndex"/> containing the merged trees and any conflicts, or null if the merge stopped early due to conflicts.
+        /// The index must be disposed by the caller.</returns>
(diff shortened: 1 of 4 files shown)


LibGit2Sharp/ObjectDatabase.cs
```diff
@@ -859,6 +848,79 @@ public virtual PackBuilderResults Pack(PackBuilderOptions options, Action<PackBu
...
+            var indexHandle = MergeCommits(ours, theirs, options, out earlyStop);
+            if (earlyStop)
+            {
+                if (indexHandle != null)
+                {
+                    indexHandle.Dispose();
+                }
+                return null;
+            }
+            var result = new Index(indexHandle, repo);
```
ethomson: I'm not wild about returning a new `Index` object whose lifecycle is tied to the `Repository`.  This is fine for services that are dealing with pull requests, since they'll spin up a repository, do a merge, then dispose the repository (and this merge result) but it's not going to be ideal for (say) GUI tools or anything that does multiple in-memory merges on the same `Repository`.

I think that we should create an `Index` subclass that's `IDisposeable` and doesn't register with the `Repository` for being disposed.  `TransientIndex`, for lack of a better name...

I briefly (and by briefly, I mean I spent way too long on it) thought about making `Index` itself disposeable and replaceable, so that you could swap in a new index on the repository.  I don't think that this is a path to success in the near term and that we should just have one type that represents the repository index and one that represents a transient index that's returned by merge / rebase / revert / etc / etc.

ethomson:
Thanks - it looks like the appveyor build is failing because of the GitHub TLS 1.2 upgrades 🙄 .  (It's the gift that keeps on giving!)

I'll push an update and I'll merge this when it's fixed.

Thanks for this work!

ethomson:
Thanks for working on this!

Issue #1550, opened 2018-03-16
Security issue with SecureUsernamePasswordCredentials 


mattjohnsonpint:
In the implementation of [`SecureUsernamePasswordCredentials`](https://github.com/libgit2/libgit2sharp/blob/d56870e14f8eacc1f0808d2075214291abf4983d/LibGit2Sharp/SecureUsernamePasswordCredentials.cs#L31), the `SecureString` is marshaled to unmanaged memory, but then is marshaled back to a regular managed `string` with the call to `Marshal.PtrToStringUni`.

Creating a managed `string` defeats the purpose of a `SecureString`.  This is discussed from a construction perspective [in the docs](https://docs.microsoft.com/en-us/dotnet/api/system.security.securestring#securestring-operations):

> **Important**
> A `SecureString` object should never be constructed from a `String`, because the sensitive data is already subject to the memory persistence consequences of the immutable `String` class. The best way to construct a `SecureString` object is from a character-at-a-time unmanaged source, such as the `Console.ReadKey` method.

Though that is describing `String` to `SecureString`, the same is also true here with `SecureString` to `String` via the `PtrToStringUni` call.

Since the native api `git_cred_userpass_plaintext_new` accepts a `char *`, we should be able to pass the data in unmanaged memory through the entire code path.

It appears we are using a custom marshaler (`StrictUtf8Marshaler`, based on `EncodingMarshaler`) here, so it would make sense (IMHO) to simply expand the implementation of that to support `SecureString` directly.   I will send a PR.

PR #1551, opened 2018-03-16
Improve SecureString marshaling 

mattjohnsonpint:
Addresses the concerns with `SecureString` usage in `SecureUsernamePasswordCredentials`, as described in #1550.

### DIFF
--- a/LibGit2Sharp/Core/EncodingMarshaler.cs
+++ b/LibGit2Sharp/Core/EncodingMarshaler.cs
@@ -2,6 +2,7 @@
 using System.Diagnostics;
 using System.Globalization;
 using System.Runtime.InteropServices;
+using System.Security;
 using System.Text;
 
 namespace LibGit2Sharp.Core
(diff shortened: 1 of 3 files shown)


ethomson:
I think that there's still a problem here - now we're allocating some heap space for the `buffer`, and we're converting the `SecureString` into UTF-8 into that buffer.  We pass that back to the native world, and we eventually [free that buffer](https://github.com/libgit2/libgit2sharp/pull/1551/files#diff-ebba9b60d30e0bb96e9266b226280f57R124).  But we do not securely zero the memory before we do.

Is that analysis correct?

mj1856:
Indeed, it seems we should zero out the memory before calling `Marshal.FreeHGlobal`.  I need to research how to best do that.  I know there is `RtlZeroMemory` macro in Windows, but I think there may be a more cross-plat friendly API we can use.   Worst case, we loop ourselves and write zeros to the same memory space, but I think that shouldn't be necessary.

bording:
@mj1856 Do you think you'll ever want to come back to this PR? Would you be okay with me closing it for now and you can re-open it if/when you make more progress?

mj1856:
@bording  - Sorry, it fell off my radar.  I'll review and get back to you shortly.

mattjohnsonpint:
I never did git back to this, sorry.  Closing for now.

mattjohnsonpint:
`SecureString` is an abomination anyway. Who knows what I was thinking. 😅

PR #1554, opened 2018-03-22
Use SetDllDirectoryinstead of SetEnvironmentVariable to set the path from where the native git lib could be found

CyberSinh:


ethomson:
Thanks for this, @CyberSinh - sorry that it took me so long to get around to taking a look.  Before I could review this, @tmat opened a PR that changes the loading behavior a bit more drastically.  Can you confirm that #1563 also will fix your problem?

ethomson:
Closed via #1563

CyberSinh:
I confirm that #1563 fix the issue.

PR #1563, opened 2018-04-12
 Avoid reading and writing global state when loading native library

tmat:
Instead of using `PATH` environment variable to locate the native DLL, call `LoadLibrary` explicitly on the path the DLL should be loaded. `DllImport` will use the already loaded DLL and won't attempt to load it again.

The reason why it is not safe to change `PATH` environment variable include race-conditions and security concerns in plugin environments, such as msbuild.
For example, if two independently developed msbuild tasks that use different versions of the libgit2 library are loaded to the same msbuild process adds path to the native library to the `PATH` environment variable there is a chance that one of the writes doesn't make it causing one of the tasks to fail to find the library.


LibGit2Sharp/Properties/AssemblyInfo.cs
```diff
@@ -16,3 +17,5 @@
 // The following GUID is for the ID of the typelib if this project is exposed to COM
 
 [assembly: Guid("c6f71967-5be1-49f5-b48e-861bff498ea3")]
+
+[assembly: InternalsVisibleTo("LibGit2Sharp.Tests, PublicKey=002400000480000094000000060200000024000052534131000400000100010013172CAC3D61EF825164EF8443ED2F97316D0C2A4A65D3B2F6E5C9175C6C589D6A0EAE803E3E7FC0DA9E6672B1DE036CF74E1D33E21DD83E1145E3A454F92E52107495082DCCD1D9F521592F79F41DF26ED727059F8A4E5D3C23ECC525306831A15F1E56B693FDE112137E973B599A13209A5B63E05EE00886DE594E70A993B5")]
```
AArnott: ⛔️ *noooooo.........*

Directory.Build.props
```diff
@@ -5,6 +5,8 @@
     <OutputPath>$(MSBuildThisFileDirectory)bin\$(MSBuildProjectName)\$(Configuration)\</OutputPath>
     <BaseIntermediateOutputPath>$(MSBuildThisFileDirectory)obj\$(MSBuildProjectName)\</BaseIntermediateOutputPath>
     <DefineConstants Condition=" '$(ExtraDefine)' != '' ">$(DefineConstants);$(ExtraDefine)</DefineConstants>
+    <SignAssembly>true</SignAssembly>
+    <AssemblyOriginatorKeyFile>..\libgit2sharp.snk</AssemblyOriginatorKeyFile>
```
AArnott: It appears you moved this from the .csproj, which is a directory deeper. But you didn't modify the relative path, so I think this is pointing at a non-existent file now.

LibGit2Sharp.TestApp/TestApp.cs
```diff
@@ -0,0 +1,46 @@
...
+            {
+                Console.Error.WriteLine("Usage: <module-name> <directory>");
+                return -1;
+            }
+
+            var moduleName = args[0];
+            var loadFromDirectory = args[1];
+            var expectedPath = Path.Combine(loadFromDirectory, (IntPtr.Size == 4) ? "x86" : "x64", moduleName + ".dll");
+
+            GlobalSettings.NativeLibraryPath = loadFromDirectory;
```
AArnott: Is it required that the NetFx host app now help libgit2sharp find the right binary now? That's unfortunate. Is there no way to avoid it? Can't the library encode the logic for how to find the DLL?

LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
```diff
@@ -2,11 +2,11 @@
 
   <PropertyGroup>
     <TargetFrameworks>net461;netcoreapp2.0</TargetFrameworks>
-    <DefineConstants Condition=" '$(TargetFramework)' == 'net461' ">$(DefineConstants);DESKTOP</DefineConstants>
   </PropertyGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\LibGit2Sharp\LibGit2Sharp.csproj" />
+    <ProjectReference Include="..\LibGit2Sharp.TestApp\LibGit2Sharp.TestApp.csproj" Condition="'$(TargetFramework)' == 'net461'" />
```
AArnott: Should you add `ReferenceOutputAssembly=false` metadata to this project reference so it doesn't proffer `Program` from the .exe into the test project?

LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
```diff
@@ -2,11 +2,11 @@
 
   <PropertyGroup>
     <TargetFrameworks>net461;netcoreapp2.0</TargetFrameworks>
-    <DefineConstants Condition=" '$(TargetFramework)' == 'net461' ">$(DefineConstants);DESKTOP</DefineConstants>
```
AArnott: Are you sure no one is using the `DESKTOP` symbol?
Why remove it in favor of using `NET461`? Doing so means if later we target net462 that all the existing `#if` directives have to be changed when semantically they just need to know if they're targeting desktop.

LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
```diff
@@ -19,8 +19,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Compile Include="..\LibGit2Sharp\Core\Epoch.cs" Link="TestHelpers\Epoch.cs" />
-    <Compile Include="..\LibGit2Sharp\Core\Platform.cs" Link="TestHelpers\Platform.cs" />
```
AArnott: Why can these be removed? Is it because you added `InternalsVisibleTo`? That seems irrelevant to this change and IMO should be omitted.
FWIW, I strongly prefer linking in the source for a few static types over using `InternalsVisibleTo`. IVT allows unbounded and unchecked access to all internals which is decidedly against the pattern these tests use so far. There is great value (in my experience) in limiting tests to public API and having IVT will make that boundary *much* harder to preserve.

LibGit2Sharp.Tests/GlobalSettingsFixture.cs
```diff
@@ -49,5 +52,35 @@ public void TryingToResetNativeLibraryPathAfterLoadedThrows()
 
             Assert.Throws<LibGit2SharpException>(() => { GlobalSettings.NativeLibraryPath = "C:/Foo"; });
         }
+
+        [ConditionalFact(typeof(NetFramework))]
```
AArnott: We already use `xunit.skippablefact` in this project. Can you use that instead of bringing in a new way to skip facts?

```cs
[SkippableFact]
public void LoadFromSpecificPath()
{
   Skip.IfNot(Platform.IsRunningOnNetFramework, ".NET Framework only test.");
   // ...
}
```

LibGit2Sharp/Properties/AssemblyInfo.cs
tmat: Why? What's the problem with IVT to tests?

Directory.Build.props
tmat: Oh, you are right. Strange it's passing. Will fix.

LibGit2Sharp.TestApp/TestApp.cs
tmat: No. it's not required. This is a test app that validates that setting `GlobalSettings.NativeLibraryPath` works.

LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
tmat: I want the .exe to be referenced by the test project. I find the .exe by `typeof(Program).Assembly.Location`.

AArnott: Ah, ok.

LibGit2Sharp/Properties/AssemblyInfo.cs
AArnott: From another comment I left:

> IVT allows unbounded and unchecked access to all internals which is decidedly against the pattern these tests use so far. There is great value (in my experience) in limiting tests to public API and having IVT will make that boundary *much* harder to preserve.

Value includes:
* focusing on user-impacting behaviors rather than internal ones
* no need to fix tests when you refactor code
* justification to fix a failing test is straightforward since it generally ties directly to a user scenario instead of some arcane internal behavior that's hard to explain to your manager.

LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
tmat: > Why can these be removed? Is it because you added InternalsVisibleTo? That seems irrelevant to this change and IMO should be omitted.

Yes. It's not irrelevant since leaving these creates duplicate types and compiler errors.

LibGit2Sharp.Tests/GlobalSettingsFixture.cs
tmat: Ok, i didn't see that.

LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
tmat: > Are you sure no one is using the DESKTOP symbol? 

Yes. Of course I searched for it. It was used on one place.

> Why remove it in favor of using NET461?

NET461 is a standard symbol. I can leave DESKTOP but imo it's better to use standard symbols rather then introducing new ones.

LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
AArnott: So it's solving a problem your IVT change created. Can we revert the IVT change instead?

Directory.Build.props
tmat: Oh, this is actually correct. Because the path is relative to the project file. I'm gonna normalize it to a full path instead.

LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
tmat: Although, it'd probably be better to not add that reference and just assume it get's copied next to the test assembly.

tmat: I meant use ReferenceOutputAssembly=false

Directory.Build.props
tmat: Yes

LibGit2Sharp.Tests/GlobalSettingsFixture.cs
```diff
@@ -53,23 +53,23 @@ public void TryingToResetNativeLibraryPathAfterLoadedThrows()
             Assert.Throws<LibGit2SharpException>(() => { GlobalSettings.NativeLibraryPath = "C:/Foo"; });
         }
 
-        [ConditionalFact(typeof(NetFramework))]
-        public void LoadFromSpecifiedPath()
+        [SkippableTheory]
+        [InlineData(new object[] { "x86" })]
+        [InlineData(new object[] { "x64" })]
```
AArnott: Nit: I think you can just say `[InlineData("x86")]` (and skip the array wrapping syntax).

tmat: You are right. Not sure why I thought it needs to be object.

bording:
@tmat That's one of the reasons why this approach makes much more sense IMHO: https://www.natemcmaster.com/blog/2017/11/11/build-tools-in-nuget/

Make it a console app that gets called at build time via targets instead.

tmat:
@bording That has bad performance.

tmat:
Unless it's a server app, but that's much more complicated than loading the native dependency. I'll continue working on this PR to make it work for Core CLR as well.

bording:
> Unless it's a server app, but that's much more complicated than loading the native dependency. I'll continue working on this PR to make it work for Core CLR as well.

Given that we need to rely on the `deps.json` file to load the correct native library, which can vary by linux distro, how are you going to mimic that? You'd need knowledge of the RID to know which library to choose.

tmat:
@bording I don't know yet. Will figure something out. Meanwhile, can we merge this PR? It's fixing it at least for .NET Framework.

bording:
@tmat I'll do a review this weekend, and I know @ethomson will likely want to take a look as well before it gets merged.

tmat:
@bording Excellent. Thanks!

bording:
> @bording I think we can keep it simple for now: the RIDs libgit2sharp.nativebinaries uses are: linux-x64, osx, win7-x64 and win7-x86. These can be easily determined from info returned by RuntimeInformation API. More can be added as needed.

That's true of the current version of LibGit2Sharp.NativeBinaries, but I'm just waiting for the RID changes I got into corefx to show up in a 2.0.x release before I can finish up https://github.com/libgit2/libgit2sharp.nativebinaries/pull/51.

The problem right now is that the single "linux" library in the NativeBinaries package only really works on Ubuntu. I want to update it to include support for every linux distro that .NET Core runs on, so that will be coming soon.

bording:
To provide a bit more info, at the very least, there's going to need to be a separate native binary for the CentOS family of distros since they use a completely different soname for their OpenSSL library.

tmat:
@bording Will it not be possible (hard) to detect the RID for the distros at runtime?

bording:
@tmat I've not spent time to determine how/if that is possible. Previously, the plan has always been to rely on .NET Core itself to figure that out vs. trying to re-implement that kind of logic directly. Aside from the MSBuild task scenario, I can't think of reason we'd want to.

tmat:
@bording MSBuild is the scenario we want it for :)

bording:
> @bording MSBuild is the scenario we want it for :)

@tmat Yeah, I realize that. 😄

My point is, aside from MSBuild, is there a reason for LibGit2Sharp to need to have RID-aware, cross-platform, manual library loading for .NET Core? I can't think of one.

To me, that feels like instead of trying to add that sort of logic here, we should be looking at adding the missing features to MSBuild instead. That way everyone benefits from the effort.

tmat:
Yes, the missing features should be added to Core CLR and MSBuild. That's hopefully gonna happen in future, but for now gitlib2sharp is unusable in msbuild tasks. I'd rather do something that's not overly complicated (and I believe this isn't) that makes it work while waiting for the real fix.

AArnott:
> for now libgit2sharp is unusable in msbuild tasks

Why do you say that, @tmat? Nerdbank.GitVersioning contains an MSBuild task that invokes libgit2sharp right now and has for many months, and runs on full MSBuild as well as MSBuild Core on all OSs.

bording:
> Yes, the missing features should be added to Core CLR and MSBuild. That's hopefully gonna happen in future, but for now libgit2sharp is unusable in msbuild tasks.

I would disagree that it's "unusable" in MSBuild tasks. People are already doing that right now. GitVerson and @AArnott's own Nerdbank.GitVersioning use it in MSBuild tasks. There's even a GitVersion PR open that's got it working under .NET Core: https://github.com/GitTools/GitVersion/pull/1269

>  I'd rather do something that's not overly complicated (and I believe this isn't) that makes it work while waiting for the real fix.

Sure, if you think that's possible without having to completely reimplement all of .NET Core's internal logic around RID parsing, etc. and still be able to handle the scenario of choosing the correct distro-specifc RID, then I'd like to see what that looks like.

tmat:
It also needs to change the `PATH` environment variable, which is not a good thing to do in msbuild process.

LibGit2Sharp/Properties/AssemblyInfo.cs
tmat: Should I instead add a new public API that returns the name of the native module, so that the tests can use it?

tmat: BTW, the tests link a couple of files from the product as a workaround for not being able to use those internal APIs. This workaround invalidates some of your points.

LibGit2Sharp/Core/Platform.cs
```diff
@@ -33,8 +33,64 @@ public static OperatingSystemType OperatingSystem
...
+                throw new PlatformNotSupportedException();
             }
         }
+
+        /// <summary>
+        /// Determines the RID to use when loading libgit2 native library.
+        /// This method only supports RIDs that are currently used by LibGit2Sharp.NativeBinaries.
+        /// </summary>
+        public static string GetNativeLibraryRuntimeId()
+        {
```
bording: To be able to support all the .NET Core linux distros, I don't want to build a library for every single one, bloating the size of the package. I want to be able to rely on the RID graph to share where it makes sense.

If I make that change and then we don't update this method, then everything except MSBuild tasks should continue to work, because .NET Core's own RID-aware library loading will find the correct library, if I understand the `dlopen/LoadLibrary` logic correctly.

Are you aware of any way get access to the RID at runtime? If not, to be able to make this method aware of the more complex RID structure that's coming, I'll first have to calculate that here also.

I really wish there was a better way to handle MSBuild's shortcomings. This is exactly the sort of code I didn't want to see in here.

version.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/AArnott/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
-  "version": "0.25.0",
+  "version": "0.26.0",
```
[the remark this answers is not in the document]
ethomson: Yes, absolutely.  We're not releasing an 0.26.0 with this.

I should have done this after releasing 0.25.0.  I'll do that now.  @tmat, please back out this change entirely.

LibGit2Sharp/Properties/AssemblyInfo.cs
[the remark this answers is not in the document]
ethomson: Agree, let's avoid `InternalsVisibleTo`.

LibGit2Sharp/Core/NativeMethods.cs
```diff
@@ -17,41 +22,64 @@ internal static class NativeMethods
...
+            if (Platform.IsRunningOnNetFramework() || Platform.IsRunningOnNetCore())
             {
-                string nativeLibraryPath = GlobalSettings.GetAndLockNativeLibraryPath();
-
-                string path = Path.Combine(nativeLibraryPath, Platform.ProcessorArchitecture);
-
-                const string pathEnvVariable = "PATH";
-                Environment.SetEnvironmentVariable(pathEnvVariable,
-                    String.Format(CultureInfo.InvariantCulture, "{0}{1}{2}", path, Path.PathSeparator, Environment.GetEnvironmentVariable(pathEnvVariable)));
+                string dllPath = Path.Combine(GlobalSettings.GetAndLockNativeLibraryPath(), libgit2 + Platform.GetNativeLibraryExtension());
```
ethomson: Bikeshedding, but let's not call it a `dllPath`.  It's only a _dll_ on Windows.  `sharedLibraryPath` or something is fully adequate.

LibGit2Sharp/Properties/AssemblyInfo.cs
ethomson: > BTW, the tests already link a couple of files from the product as a workaround for not being able to use those internal APIs.

That's unfortunate.  Can you raise an issue pointing to those?

Directory.Build.props
```diff
@@ -5,6 +5,8 @@
     <OutputPath>$(MSBuildThisFileDirectory)bin\$(MSBuildProjectName)\$(Configuration)\</OutputPath>
     <BaseIntermediateOutputPath>$(MSBuildThisFileDirectory)obj\$(MSBuildProjectName)\</BaseIntermediateOutputPath>
     <DefineConstants Condition=" '$(ExtraDefine)' != '' ">$(DefineConstants);$(ExtraDefine)</DefineConstants>
+    <SignAssembly>true</SignAssembly>
+    <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)libgit2sharp.snk</AssemblyOriginatorKeyFile>
```
ethomson: Why are we signing this now?

LibGit2Sharp.TestApp/TestApp.cs
ethomson: So, at the very least, `TestApp` is not a very descriptive name.  (What does it test?  Is it actually running as part of our test gauntlet?)

Let's at least give it a better name.  But having it actually _run_ as part of the tests would be much more useful.  If we can't do that, then I'm not sure we should be checking it in at all.

ethomson: Ugh, sorry.  You're consistently a step ahead of me.  I do now see the test executing this.

Still, let's improve the name here a bit.  `NativeLoaderTest` or something?  If we add additional apps, it will be useful to disambiguate.

LibGit2Sharp/Core/Platform.cs
```diff
@@ -33,8 +33,64 @@ public static OperatingSystemType OperatingSystem
...
+            switch (OperatingSystem)
+            {
+                case OperatingSystemType.MacOSX:
+                    return "osx";
+
+                case OperatingSystemType.Unix:
+                    return "linux-" + ProcessorArchitecture;
```
ethomson: Please don't do this: Unix != Linux.  If we need to have a constant that means linux, instead of a generic Unix constant, then let's name it _Linux_.

LibGit2Sharp/Core/NativeMethods.cs
```diff
@@ -17,41 +22,64 @@ internal static class NativeMethods
...
+            }
+            finally // avoid thread aborts
+            {
+                // Initialization can be called multiple times as long as there is a corresponding shutdown to each initialization.
+                initCounter = git_libgit2_init();
+                shutdownObject = new NativeShutdownObject();
+            }
+
+            // Configure the OpenSSL locking on the first initialization of the library in the current process.
+            if (initCounter == 1)
```
ethomson: Help me better understand this construct?

What happens if a thread aborts immediately _before_ the `try/finally` block is executed?  In that case `git_libgit2_init` will _not_ have been called, but the `NativeShutdownObject` finalizer will now run since it's now a `CriticalFinalizerObject`...

If that analysis is correct then we need to protect against this.

ethomson:
Thanks for doing this work, @tmat.  This definitely seems like it will be an improvement over the `PATH` based hackery that we had been doing.  Big thanks to @bording and @AArnott for the review; I had just a few minor comments to add beyond what they'd said.

LibGit2Sharp/Properties/AssemblyInfo.cs
AArnott: Linking a few source files with static types in both product and test assemblies isn't ideal, but in my experience it is _far_ preferable to IVT because (as you point out) it severely limits the scope so we don't end up with half our tests written against internals.

Directory.Build.props
AArnott: @tmat had to strong-name sign the test assembly in order to make InternalsVisibleTo available to it. When he removes that, he can revert the signing change too.

LibGit2Sharp/Properties/AssemblyInfo.cs
tmat: OK, I'll try to make it work without IVT.

LibGit2Sharp.TestApp/TestApp.cs
tmat: I'll rename it to `NativeLoaderTestApp`. It is used by the new test I added. To test library loading we need to spawn a process, the test app is used for that.

version.json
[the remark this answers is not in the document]
tmat: Sure, I'll revert this file. BTW, when can you release bits with this PR included?

LibGit2Sharp/Core/NativeMethods.cs
```diff
@@ -17,41 +22,64 @@ internal static class NativeMethods
...
-                string path = Path.Combine(nativeLibraryPath, Platform.ProcessorArchitecture);
-
-                const string pathEnvVariable = "PATH";
-                Environment.SetEnvironmentVariable(pathEnvVariable,
-                    String.Format(CultureInfo.InvariantCulture, "{0}{1}{2}", path, Path.PathSeparator, Environment.GetEnvironmentVariable(pathEnvVariable)));
+                string dllPath = Path.Combine(GlobalSettings.GetAndLockNativeLibraryPath(), libgit2 + Platform.GetNativeLibraryExtension());
+
+                // Try to load the .dll from the path explicitly.
+                // If this call succeeds further DllImports will find the library loaded and not attempt to load it again.
+                // If it fails the next DllImport will load the library from safe directories.
```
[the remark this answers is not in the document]
tmat: I'm gonna change the logic a bit and allow the consumer of the library to skip this step entirely to address the scenario you mentioned.

LibGit2Sharp/Core/NativeMethods.cs
tmat: Fixed.

LibGit2Sharp/Core/NativeMethods.cs
tmat: In that case ```NativeShutdownObject``` finalizer will not be called because the object hasn't been created yet.

The only purpose of the try-finally block is to make these two lines uninterruptible by thread abort:
```C#
                initCounter = git_libgit2_init();
                shutdownObject = new NativeShutdownObject();
```

LibGit2Sharp/Core/Platform.cs
tmat: Removed. The msbuild tasks will need to set NativeLibraryPath themselves now.

LibGit2Sharp/Core/NativeMethods.cs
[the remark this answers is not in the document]
tmat: Done.

LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
```diff
@@ -18,10 +20,20 @@
...
+      <_TestAppFile Include="@(TestAppExe->'%(RootDir)%(Directory)%(Filename).exe')" />
+      <_TestAppFile Include="@(TestAppExe->'%(RootDir)%(Directory)%(Filename).exe.config')" />
+      <_TestAppFile Include="@(TestAppExe->'%(RootDir)%(Directory)%(Filename).pdb')" />
+    </ItemGroup>
+      
+    <ItemGroup>
+      <Content Include="@(_TestAppFile)" CopyToOutputDirectory="PreserveNewest" Visible="false" />
+    </ItemGroup>
+  </Target>
+
```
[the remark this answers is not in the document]
tmat: Fixed.

LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
```diff
@@ -18,10 +20,20 @@
...
-    <Compile Include="..\LibGit2Sharp\Core\Epoch.cs" Link="TestHelpers\Epoch.cs" />
-    <Compile Include="..\LibGit2Sharp\Core\Platform.cs" Link="TestHelpers\Platform.cs" />
-    <Compile Remove="desktop\**" Condition=" '$(TargetFramework)' != 'net461' " />
+  <Target Name="CopyTestAppExes" AfterTargets="ResolveProjectReferences">
+    <ItemGroup>
+      <_TestAppFile Include="@(TestAppExe->'%(RootDir)%(Directory)%(Filename).exe')" />
+      <_TestAppFile Include="@(TestAppExe->'%(RootDir)%(Directory)%(Filename).exe.config')" />
+      <_TestAppFile Include="@(TestAppExe->'%(RootDir)%(Directory)%(Filename).pdb')" />
+    </ItemGroup>
+
```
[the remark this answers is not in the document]
tmat: Fixed.

LibGit2Sharp/LibGit2Sharp.csproj
```diff
@@ -9,9 +9,7 @@
     <PackageTags>libgit2 git</PackageTags>
     <PackageProjectUrl>https://github.com/libgit2/libgit2sharp/</PackageProjectUrl>
     <Authors>LibGit2Sharp contributors</Authors>
-    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
-    <SignAssembly>true</SignAssembly>
-    <AssemblyOriginatorKeyFile>..\libgit2sharp.snk</AssemblyOriginatorKeyFile>
+    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
```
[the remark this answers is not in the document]
tmat: Fixed.

LibGit2Sharp/Properties/AssemblyInfo.cs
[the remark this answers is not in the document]
tmat: Done.

BTW, I noticed the unique id generator is generating a new guid every build. That's unfortuante since it makes the entire build non-deterministic. The guid seems to be only passed to the custom marshaller as a cookie, which doesn't use it for anything. So it seems unnecessary.

tmat:
I believe I addressed all feedback now.

LibGit2Sharp/Core/NativeMethods.cs
ethomson: Oh, duh.  You're of course right.  I forgot that was the entire purpose of `NativeShutdownObject`.  🙄

LibGit2Sharp/Properties/AssemblyInfo.cs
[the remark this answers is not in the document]
ethomson: Right; if you wanted to send up a PR for that I'd be happy to 👀 .  Let's not do it here though, so we can get this 🚢 for you.

ethomson:
@AArnott, @bording do you have any final thoughts on this?

LibGit2Sharp/Properties/AssemblyInfo.cs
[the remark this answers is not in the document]
tmat: Yeah, I'm not doing it in this PR for sure.

bording:
@ethomson I'd like to do another review pass to take a look at the new changes. Should be able to get to that later today.

bording:
@tmat Allowing the `NativeLibraryPath` to work with .NET Core but not have a default value seems like a good middle ground, though that does mean the "which library should I choose based on my RID?" problem is now something that has to be solved by the caller.

Also, regarding:
>BTW, I noticed the unique id generator is generating a new guid every build. That's unfortuante since it makes the entire build non-deterministic. The guid seems to be only passed to the custom marshaller as a cookie, which doesn't use it for anything. So it seems unnecessary.

That does at least have a historical purpose. It's there to work around a CLR bug apparently: https://github.com/libgit2/libgit2sharp/pull/438#issue-6003591. If that caching bug has been fixed for all the frameworks that we run on now, then it seems like it could be removed.

tmat:
> though that does mean the "which library should I choose based on my RID?" problem is now something that has to be solved by the caller.

Yes, that's why I originally put it to this PR.

tmat:
@bording  Thanks for review!

bording:
> Yes, that's why I originally put it to this PR.

Are you aware of any way to determine the RID at runtime other than having to construct it manually the same way the host does? If that was possible, then it would be more feasible to have a reliable default value for .NET Core.

tmat:
> That does at least have a historical purpose. It's there to work around a CLR bug apparently: #438 (comment). If that caching bug has been fixed for all the frameworks that we run on now, then it seems like it could be removed.

Could the MarshalCookie be the commit sha insted of a random guid?

tmat:
I don't think there is public API. There is this:
https://github.com/dotnet/core-setup/blob/master/src/managed/Microsoft.DotNet.PlatformAbstractions/RuntimeEnvironment.cs#L15-L17

tmat:
https://github.com/dotnet/corefx/issues/16629

bording:
> Could the MarshalCookie be the commit sha insted of a random guid?

It seems like that would work, at least for actually shipped assemblies. You could still run into a problem if you loaded two different versions of the assembly that you built locally, though.

However, what I'd really like to know is if it's even still a problem any more. It was mentioned in https://github.com/libgit2/libgit2sharp/pull/430#issuecomment-18375744 that it was filed as a CLR bug. Do you by any chance have a way to find out if it ever got fixed?

eerhardt:
> I don't think there is public API for RID currently. There is this:
https://github.com/dotnet/core-setup/blob/master/src/managed/Microsoft.DotNet.PlatformAbstractions/RuntimeEnvironment.cs

That is the public API currently.  You can get it here:  https://www.nuget.org/packages/Microsoft.DotNet.PlatformAbstractions/

AArnott:
> > Could the MarshalCookie be the commit sha insted of a random guid?

> It seems like that would work, at least for actually shipped assemblies. You could still run into a problem if you loaded two different versions of the assembly that you built locally, though.

Even two versions built locally would have different commit IDs, except when you're building uncommitted changes multiple times, but how would that ever be a problem in practice?

Since you're using Nerdbank.GitVersioning, you should be able to get the commit ID from a build property, or a unique version from the `ThisAssembly.AssemblyInformationalVersion` at runtime.

ethomson:
Thanks again, everyone, this is now published at https://www.nuget.org/packages/LibGit2Sharp/0.26.0-preview-0017

PR #1571, opened 2018-04-22
[WIP] Resolve native library location from runtimes folder

bording:
This is the LibGit2Sharp side of the changes that https://github.com/libgit2/libgit2sharp.nativebinaries/pull/68 enables, which means it won't build until an update package is available.

This builds on top of the changes introduced in #1563, allowing the default native library path to be resolved from the runtimes folder, for all target frameworks (including mono).

I've embedded the runtime.json information from https://github.com/dotnet/corefx/blob/master/pkg/Microsoft.NETCore.Platforms/runtime.json, and added the Microsoft.DotNet.PlatformAbstractions package.

This lets the correct native library to be chosen based on the runtime identifier, which maps back to the OS that the code is being run on.

One immediate benefit from this change is that when using LibGit2Sharp on .NET Core inside an MSBuild task, the correct native library should be chose automatically.

@tmat @AArnott I think this should really simplify the MSBuild task scenario while we're waiting for MSBuild to support this directly!


bording:
I know that there are some tests that will fail because of these new changes, so I'll be sure to fix those up once we get a new native binaries package on nuget.org.

ethomson:
Sorry, looks like my long delay allowed some conflicts to creep in.  :(  Happy to merge this if you wouldn't mind fixing them up.

PR #1602, opened 2018-08-25
Package updates

bording:
Also includes a minor tweak to the editorconfig file.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,9 +10,9 @@ trim_trailing_whitespace = true
 insert_final_newline = true
 
 ; Not change VS generated files
-[*.{sln,csroj}]
+[*.{sln,csproj}]
 trim_trailing_whitespace = false
 insert_final_newline = false
 
-[*.{props,targets}]
+[*.{props,targets,csproj}]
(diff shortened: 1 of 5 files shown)


ethomson:
With the travis update to require a newer .NET Core, am I right in thinking that we are now no longer .NET Core 2.0 compatible?  What's pushing that requirement?  What is the fallout from that?

bording:
The reason for that is to get a newer SDK, which is needed for the switch to Microsoft.SourceLink.GitHub. 

That has no impact on the compatibility of the library. That's determined by the TFMs we target, which hasn't changed. The library is still `netstandard2.0`.

bording:
It appears that all of the AppVeyor builds on the libgit2 account are getting stuck at being queued, but never actually run.

I've opened https://help.appveyor.com/discussions/problems/16035-queued-builds-never-start, so hopefully they'll be able to fix the problem.

bording:
AppVeyor fixed the problem with the build queue!

ethomson:
Yep, makes sense.  Thanks @bording.

PR #1604, opened 2018-09-05
Update Readme Prerequisites .NET/Mono versions

devlead:
Library now targets .NETStandard 2.0 which bumps .NET to 4.6.1+ and Mono to 5.4+
https://docs.microsoft.com/en-us/dotnet/standard/net-standard

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,8 +22,8 @@
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


ethomson:
Thanks @devlead!

PR #1606, opened 2018-09-09
Provide .NET 4.6.1 framework target

### LINKED ISSUE #1605: Consider providing a net461 moniker in nuget package
As per resent guidance provided by @terrajobst on Twitter [recently](https://twitter.com/terrajobst/status/1031999730320986112)

Main take aways
> If you want to consume .NET Standard 1.5+ from .NET Framework, I recommend to be on 4.7.2.
> If you dual target for .NET Standard and .NET Framework, you’ll get the best of both worlds.

Issue as a library author is many assemblies has dropped the .NET Framework monikers, libgit2sharp included, which means if I have libgit2sharp as a dependency, I myself can't target 4.6.1.

Did a quick PoC win fork [devlead/libgit2sharp/cf417d65a2e2d8192e9b2676794372575d6ae49a](https://github.com/devlead/libgit2sharp/blob/cf417d65a2e2d8192e9b2676794372575d6ae49a/LibGit2Sharp/LibGit2Sharp.csproj), things seems to build & test locally.

Basic changes I did

https://github.com/devlead/libgit2sharp/blob/cf417d65a2e2d8192e9b2676794372575d6ae49a/LibGit2Sharp/LibGit2Sharp.csproj#L4
```xml
    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
```

https://github.com/devlead/libgit2sharp/blob/cf417d65a2e2d8192e9b2676794372575d6ae49a/LibGit2Sharp/LibGit2Sharp.csproj#L41-L43
```xml
 <ItemGroup Condition=" '$(TargetFramework)' == 'net461' ">
    <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" PrivateAssets="none" />
  </ItemGroup>
```

If this is something you might consider, I'll happily send in this as a PR.

devlead:
As discuessed in #1605
* Adds .NET 4.6.1 framework target
* adds Pollyfill for System.Runtime.InteropServices.RuntimeInformation
* fixes #1605

### DIFF
--- a/LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
+++ b/LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
@@ -28,7 +28,6 @@
   <Target Name="CopyTestAppExes" AfterTargets="ResolveProjectReferences">
     <ItemGroup>
       <_TestAppFile Include="@(TestAppExe->'%(RootDir)%(Directory)%(Filename).exe')" />
-      <_TestAppFile Include="@(TestAppExe->'%(RootDir)%(Directory)%(Filename).exe.config')" />
       <_TestAppFile Include="@(TestAppExe->'%(RootDir)%(Directory)%(Filename).pdb')" />
     </ItemGroup>
 
--- a/LibGit2Sharp/Core/Platform.cs
+++ b/LibGit2Sharp/Core/Platform.cs
@@ -13,11 +13,30 @@ namespace LibGit2Sharp.Core
     internal static class Platform
     {
         public static string ProcessorArchitecture => IntPtr.Size == 8 ? "x64" : "x86";
+#if NETFRAMEWORK 
+        private static bool? _isRunningOnMac;
+        private static bool IsRunningOnMac() => _isRunningOnMac ?? (_isRunningOnMac = TryGetIsRunningOnMac()) ?? false;
+#endif
 
         public static OperatingSystemType OperatingSystem
         {
             get
             {
+#if NETFRAMEWORK
+                var platform = (int)Environment.OSVersion.Platform;
+                if (platform <= 3 || platform == 5)
+                {
+                    return OperatingSystemType.Windows;
+                }
+                if (IsRunningOnMac())
+                {
+                    return OperatingSystemType.MacOSX;
+                }
+                if (platform == 4 || platform == 6 || platform == 128)
+                {
+                    return OperatingSystemType.Unix;
+                }
+#else
                 if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                 {
                     return OperatingSystemType.Windows;
@@ -32,7 +51,7 @@ namespace LibGit2Sharp.Core
                 {
                     return OperatingSystemType.MacOSX;
                 }
-
+#endif
                 throw new PlatformNotSupportedException();
             }
         }
@@ -71,5 +90,70 @@ namespace LibGit2Sharp.Core
         /// </summary>
         public static bool IsRunningOnNetCore()
             => typeof(object).Assembly.GetName().Name != "mscorlib";
+
+#if NETFRAMEWORK
+#pragma warning disable IDE1006 // Naming Styles
+        [DllImport("libc")]
+        private static extern int sysctlbyname(
+            [MarshalAs(UnmanagedType.LPStr)] string property,
+            IntPtr output,
+            IntPtr oldLen,
+            IntPtr newp,
+            uint newlen);
+#pragma warning restore IDE1006 // Naming Styles
+
+        private static bool TryGetIsRunningOnMac()
+        {
+            const string OsType = "kern.ostype";
+            const string MacOsType = "Darwin";
+
+            return MacOsType == GetOsType();
+
+            string GetOsType()
+            {
+                try
+                {
+                    IntPtr
+                        pointerLength = IntPtr.Zero,
+                        pointerString = IntPtr.Zero;
+
+                    try
+                    {
+                        pointerLength = Marshal.AllocHGlobal(sizeof(int));
+
+                        sysctlbyname(OsType, IntPtr.Zero, pointerLength, IntPtr.Zero, 0);
+
+                        var length = Marshal.ReadInt32(pointerLength);
+
+                        if (length <= 0)
+                        {
+                            return string.Empty;
+                        }
+
+                        pointerString = Marshal.AllocHGlobal(length);
+
+                        sysctlbyname(OsType, pointerString, pointerLength, IntPtr.Zero, 0);
+
+                        return Marshal.PtrToStringAnsi(pointerString);
+                    }
+                    finally
+                    {
+                        if (pointerLength != IntPtr.Zero)
+                        {
+                            Marshal.FreeHGlobal(pointerLength);
+                        }
+                        if (pointerString != IntPtr.Zero)
+                        {
+                            Marshal.FreeHGlobal(pointerString);
+                        }
+                    }
+                }
+                catch
+                {
+                    return null;
+                }
+            }
+        }
+#endif
     }
 }
--- a/LibGit2Sharp/Core/NativeMethods.cs
+++ b/LibGit2Sharp/Core/NativeMethods.cs
@@ -37,7 +37,11 @@ namespace LibGit2Sharp.Core
                     // Try to load the .dll from the path explicitly.
                     // If this call succeeds further DllImports will find the library loaded and not attempt to load it again.
                     // If it fails the next DllImport will load the library from safe directories.
+#if NETFRAMEWORK
+                    if (Platform.OperatingSystem == OperatingSystemType.Windows)
+#else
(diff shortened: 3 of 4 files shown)


LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
```diff
@@ -29,7 +29,6 @@
   <Target Name="CopyTestAppExes" AfterTargets="ResolveProjectReferences">
     <ItemGroup>
       <_TestAppFile Include="@(TestAppExe->'%(RootDir)%(Directory)%(Filename).exe')" />
-      <_TestAppFile Include="@(TestAppExe->'%(RootDir)%(Directory)%(Filename).exe.config')" />
```
devlead: This I can revert, seems cause issues when on a feature branch.

devlead:
It would seem MSBuild property `libgit2_hash` isn't populated when I change from `<TargetFramework>netstandard2.0</TargetFramework>` to `<TargetFrameworks>netstandard2.0;net461</TargetFrameworks>`. Will see if I can see where it's set.

jrgcubano:
@devlead wooo.. It would be wonderful if you could get this... I'm trying to migrate https://github.com/YellowLineParking/Appy.GitDb and we want to keep both frameworks too.

devlead:
Ok checking MSBuild log seems `LibGit2Sharp.NativeBinaries.props` isn't imported so that explains why `libgit2_hash` isn't populated could be related to props in `build` and not `buildCrossTargeting` will see if I can work around that.

bording:
@devlead Yeah, since this PR adds multi-targeting, that will be an issue. We'll have to update LibGit2Sharp.NativeBinaries first before we can move forward with this.

devlead:
@bording tested with package generated from libgit2/libgit2sharp.nativebinaries#75

https://ci.appveyor.com/project/devlead/libgit2sharp-nativebinaries/build/1.0.1

And then I could successfully build a nuget package.

ethomson:
Neat!  Once the NativeBinaries changes are ready to go, we'll merge this.  Thanks!

bording:
@ethomson Just an FYI I do have some review comments to make before we merge this, but I was waiting for the NativeBinaries update.

devlead:
@bording I have now updated libgit2/libgit2sharp.nativebinaries#75 to import and not copy.
I could add a temporary nuget.config pointing to my MyGet feed if you would like this to build on CI and review before native binaries are shipped, and I can revert and rebase when they have.

bording:
https://github.com/libgit2/libgit2sharp.nativebinaries/pull/75 has been released as LibGit2Sharp.NativeBinaries 1.0.233, so you can update this PR to use it.

devlead:
@bording native binaries updated and PR rebased.

devlead:
> 2018-09-17T22:02:17.4868779Z ##[section]Async Command Start: Update Build Number
> 2018-09-17T22:02:17.4892551Z ##[section]Async Command End: Update Build Number

Azure Pipelines currently doesn't allow setting build numbers on PRs, whish thet did, but that's why builds are failing. But Mac / Linux `Test Run Successful.`  Windows still running.

bording:
> @bording native binaries updated and PR rebased.

Cool, I should be able to look over this tomorrow then.

LibGit2Sharp/Core/Platform.cs
```diff
@@ -71,5 +90,45 @@ public static bool IsRunningOnNetFramework()
         /// </summary>
         public static bool IsRunningOnNetCore()
             => typeof(object).Assembly.GetName().Name != "mscorlib";
+
+#if !NETCORE
+#pragma warning disable IDE1006 // Naming Styles
+        [DllImport("libc")]
```
bording: Is there a danger here of this throwing? For example, what if we're running on a musl libc instead of glibc?

ethomson: Unlikely to find a system without libc (even if it’s not glibc) but this is not ideal.  `uname` takes a `struct` of char arrays (of a well defined size) not a single char array.  So this should work everywhere (since it’s an appropriately large buffer) but I don’t love it.

I don’t think that libgit2 has a mechanism to tell you what OS you’re running on, but it seems like we could.

But without that, I think I would recommend `sysctl` instead of `uname`.

Directory.Build.props
```diff
@@ -7,4 +7,8 @@
     <DefineConstants Condition=" '$(ExtraDefine)' != '' ">$(DefineConstants);$(ExtraDefine)</DefineConstants>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
+    <DefineConstants>$(DefineConstants);NETCORE</DefineConstants>
```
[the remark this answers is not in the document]
devlead: Reverted.

LibGit2Sharp/Core/Platform.cs
```diff
@@ -13,11 +13,16 @@ internal enum OperatingSystemType
     internal static class Platform
     {
         public static string ProcessorArchitecture => IntPtr.Size == 8 ? "x64" : "x86";
+#if !NETCORE
+        private static bool? _isRunningOnMac;
+        public static bool IsRunningOnMac() => _isRunningOnMac ?? (_isRunningOnMac = TryGetIsRunningOnMac()) ?? false;
```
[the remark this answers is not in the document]
devlead: Internal class, but sure made it private.

LibGit2Sharp/Core/Platform.cs
devlead: > Is there a danger here of this throwing? For example, what if we're running on a musl libc instead of glibc?

`libc` should not be an issue.

> I don’t think that libgit2 has a mechanism to tell you what OS you’re running on, but it seems like we could.

As this is used to decide how to load `libgit`, I don't see how we could use it, right?

> But without that, I think I would recommend sysctl instead of uname.

Used `uname` trick for years in Mono projects, so didn't think of `sysctl` will investigate.

devlead: Haven't looked at  `sysctl` implementation yet, but while seeking inspiration it does seem many use `uname` for os detection and then `sysctl` for kernel revision detection, wonder if there's some reasoning around that.
https://github.com/dotnet/corefx/blob/0566028e42a8f2fa86b8e0dd856d1f13d0fdce05/src/Native/Unix/System.Native/pal_runtimeinformation.c#L25-L39

Could be a legacy from the project "k" days, which I now see also used uname
https://github.com/aspnet/PlatformAbstractions/blob/174f51f3180ac9718de48d155b819e8a05310ab0/src/Microsoft.Extensions.PlatformAbstractions/Native/NativeMethods.Unix.cs#L10-L27

Maybe  `sysctl` `kern.ostype` is good enough will give it a go and refactor PR.

devlead: Refactored PR to use `sysctl` and  `kern.ostype`.

devlead:
@bording @ethomson think I've addressed all feedback and PR ready for review again.

LibGit2Sharp/Core/Platform.cs
```diff
@@ -13,11 +13,16 @@ internal enum OperatingSystemType
     internal static class Platform
     {
         public static string ProcessorArchitecture => IntPtr.Size == 8 ? "x64" : "x86";
+#if !NETSTANDARD
```
bording: I want to change this eventually, but Platform.cs is also included in the Tests project, so using `NETSTANDARD` here means the test project has these members defined when they don't need to be.

To make this work in both places, basing the logic on `NETFRAMEWORK` is going to be the way to go.

devlead: Addressed

jrgcubano:
@bording ready for next release? 🙄

ethomson:
Thanks @devlead for doing this, and thanks @bording for the review.

devlead:
@bording @ethomson splendid 👍

AArnott:
Can we get a new release to nuget.org with this fix please?

devlead:
Interesting the GH release 0.25.3 has this PR in it
https://github.com/libgit2/libgit2sharp/compare/v0.25.3...master

But the NuGet release seems to have been done from
https://github.com/libgit2/libgit2sharp/commits/ethomson/release_0_25_3

Which doesn't have this pr in it, so currently the release doesn't match the release binaries 🤔 

But perhaps that was unintentional and this will come in 0.26.0?

devlead:
Ok I just clicked commits on the release 
https://github.com/libgit2/libgit2sharp/releases/tag/v0.25.3

AArnott:
That's a misleading link, I guess. But look carefully at the wording of that link:

>  63 commits to **master** *since* this release

devlead:
Ah sorry you're correct.

ethomson:
> Can we get a new release to nuget.org with this fix please?

@AArnott: you bet - https://www.nuget.org/packages/LibGit2Sharp/0.26.0-preview-0054

PR #1616, opened 2018-10-07
Update NativeBinaries to 1.0.235

ethomson:


bording:
Bumping up to 235 here does mean that the 0.25 release will get a NativeBinaries version that has the more expanded native binary set, which has previously only been in the 0.26 prereleases.

Not sure that's a problem, just wanted to point out the change.

ethomson:
It will - and the underlying LibGit2Sharp here doesn’t know anything about them, so many of the binaries are largely useless and will never get loaded by anyone.  I’m disappointed that we’ll be bloating the nuget restore process for people for no benefit but I think I’m happier with that than trying to create another new (but different) native binaries package for the libgit2 point release.

ethomson:
> That isn't true. 0.25.x targets netstandard2.0, so the extra binaries will be used in the same way that they are in 0.26.x.

Oh, cool.  I thought that #1571 was required for the multiple platforms.  Nice to know that the extra MBs aren't for naught.

PR #1628, opened 2018-11-02
Fix buildandtest.sh to use - instead of /

lupino3:
The '/' character will be interpreted as a path start under Linux (and
WSL), so change it to a '-'.

Tested locally under WSL (Windows Subsystem for Linux).


ethomson:
👍

PR #1630, opened 2018-11-02
Fix resources path computation in test setup

### LINKED ISSUE #1629: Tests don't run from command-line or VSCode
### Reproduction steps
1. Clone the repository from the WSL command line
2. Open the folder in Visual Studio Code
3. Try to run one test (click on "Run Test" over any test)

Alternatively, after cloning, run `dotnet restore` and run any test with `dotnet test`, for example:

```
$ dotnet test LibGit2Sharp.Tests//LibGit2Sharp.Tests.csproj --filter LibGit2Sharp.Tests.UnstageFixture.CanUnstageSourceOfARename
```

### Expected behavior
The test runs, and either passes or fails.

### Actual behavior
The test always fails, cannot open the `testrepo_wd` directory. Example:

```
System.IO.DirectoryNotFoundException : Could not find a part of the path 'C:\Users\<my username>\source\github\LibGit2Sharp.Tests\Resources\testrepo_wd'.
```

The directory is not found because the path is wrong. It is missing the git repo name. The correct one would be:

```
C:\Users\<my username>\source\github\libgit2sharp\LibGit2Sharp.Tests\Resources\testrepo_wd
```

Note that this is fixed if the environment variable `Configuration` is set to `release`, for some reason.

0a9c84dd98cb6b322d63009b4ab85e728f38317c (current HEAD)

Windows 10 Enterprise 1803
.NET Core 2.1.403

lupino3:
The resources path is set based on the path of the current assembly.

The problem is that the assembly directory has a different number of
components, depending on whether the Configuration environment variable
is set or not.

Therefore, the previous code, that depended on the number of subdirs,
would fail if there is no value for the Configuration environment
variable.

The new code looks for the occurrence of a well-known string in the path
and bases the path of the test resources off that position.

It is still pretty sub-optimal, but I think it's a bit less sub-optimal
than it was.

Tested in the following ways:
- run dotnet test on one test from the command line on Windows;
- click on "Run Test" from Visual Studio Code.

Fixes #1629.

### DIFF
--- a/LibGit2Sharp.Tests/TestHelpers/BaseFixture.cs
+++ b/LibGit2Sharp.Tests/TestHelpers/BaseFixture.cs
@@ -64,9 +64,8 @@ namespace LibGit2Sharp.Tests.TestHelpers
             if (resourcesPath == null)
             {
                 string initialAssemblyParentFolder = Directory.GetParent(new Uri(typeof(BaseFixture).GetTypeInfo().Assembly.CodeBase).LocalPath).FullName;
-                const string sourceRelativePath = @"../../../../LibGit2Sharp.Tests/Resources";
-
-                resourcesPath = Path.Combine(initialAssemblyParentFolder, sourceRelativePath);
+                int pos = initialAssemblyParentFolder.IndexOf("LibGit2Sharp.Tests");
+                resourcesPath = Path.Combine(initialAssemblyParentFolder.Substring(0, pos), "../LibGit2Sharp.Tests/Resources");
             }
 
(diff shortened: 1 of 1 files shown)


ethomson:
As you noted, this is imperfect - we might want to consider copying the entire `Resources` directory to the output dir so that we _know_ where it is.  I was hoping that's as simple as a few lines of msbuild, but it does not seem to be.

Anyway, I'm not in love with this but this is a strict improvement over what's there now, which I'm surprised _ever_ worked.  :D  

Thanks for this!

PR #1632, opened 2018-11-02
Tests: copy test resources to output directory

ethomson:
Copy the test resources to the output directory so that we can resolve the easily, instead of trying to resolve them from the source directory.

This is a refinement on #1630 (with help from @b1thunt3r).


lupino3:
Much cleaner, thanks!

PR #1618, opened 2018-10-08
WIP: Use managed HTTP stack

ethomson:
In an attempt to reduce our native dependencies, we can use the managed HTTP stack for connecting to HTTP remotes.  This is somewhat working but lacking a number of pieces of functionality:

- [x] ~~Proxies~~
- [x] Default credentials
- [x] ~~Certificate validation callbacks~~
- [x] Custom user agents
- [x] Overriding with a different custom transport

However, for users who do not do any networking, this may be worth evaluation.

PR #1636, opened 2018-11-03
Update LibGit2Sharp.NativeBinaries to 1.0.245

bording:
The new version of LibGit2Sharp.NativeBinaries contains the following changes:
- There is a new `ubuntu.18.04-x64` binary
- The `rhel-x64` binary is now built with glibc 2.12, so it should work with RHEL6

This should help stop some of the pain people have been having while we continue to work on the real solution: #1618


bording:
@ethomson I'm going to go ahead and merge this and push up a new preview since I know there are people blocked by this.

PR #1639, opened 2018-11-06
Update NativeBinaries to libgit2 0.27.7

ethomson:


ethomson:
**sigh**

ethomson:
I'm going to hold off on merging this until I do track down the root cause tonight, in case it's a behavior change that we think should be addressed in libgit2.

ethomson:
Yep, was a change in behavior that I accidentally introduced in libgit2 and made it into 0.27.7.  Fixing that change over there, then I'll do a proper native binaries build.  Oops.

PR #1645, opened 2018-11-29
Update NativeBinaries to 1.0.258

ethomson:
No more zlib or libcurl dependencies!


bording:
Hmm, test failures. That's unexpected!

bording:
> In particular there were some changes around config settings that I think renders this test obsolete.

Ok that's good. I was thinking that somehow the zlib removal was responsible!

ethomson:
libgit2 changed the `git_config_entry` structure, we just needed to update ours to match.

PR #1651, opened 2018-12-19
Target net46 instead of net461

bording:
In order to better support using LibGit2Sharp from inside of Visual Studio, we need to switch to `net46` instead of `net461`. See https://github.com/github/VisualStudio/issues/1849#issuecomment-411570902 for details.


bording:
CC: @jcansdale

PR #1652, opened 2018-12-19
Add support for git diff --indent-heuristic option

jcansdale:
### What this PR does

- Add `IndentHeuristic` option to `CompareOptions` (it's a `bool`)
- Add test to check that `--indent-heuristic` makes a difference
- Add test for when `--indent-heuristic` won't make a difference

### DIFF
--- a/LibGit2Sharp.Tests/DiffBlobToBlobFixture.cs
+++ b/LibGit2Sharp.Tests/DiffBlobToBlobFixture.cs
@@ -128,7 +128,7 @@ namespace LibGit2Sharp.Tests
         }
 
         [Fact]
-        public void ComparingBlobsWithIndentHeuristicOptionMakesADifference()
+        public void ComparingBlobsWithNoSpacesAndIndentHeuristicOptionMakesADifference()
         {
             var path = SandboxStandardTestRepoGitDir();
             using (var repo = new Repository(path))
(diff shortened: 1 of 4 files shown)


LibGit2Sharp.Tests/DiffBlobToBlobFixture.cs
```diff
@@ -126,5 +126,79 @@ public void ComparingTwoNullBlobsReturnsAnEmptyContentChanges()
...
+	4";
+                var oldBlob = repo.ObjectDatabase.CreateBlob(new MemoryStream(Encoding.UTF8.GetBytes(oldContent)));
+                var newBlob = repo.ObjectDatabase.CreateBlob(new MemoryStream(Encoding.UTF8.GetBytes(newContent)));
+                var noIndentHeuristicOption = new CompareOptions { IndentHeuristic = false };
+                var indentHeuristicOption = new CompareOptions { IndentHeuristic = true };
+
+                ContentChanges changes0 = repo.Diff.Compare(oldBlob, newBlob, noIndentHeuristicOption);
+                ContentChanges changes1 = repo.Diff.Compare(oldBlob, newBlob, indentHeuristicOption);
+
+                Assert.NotEqual(changes0.Patch, changes1.Patch);
```
Therzok: Can we validate that changes0 is a superset of changes1?

Therzok: Or validate the results of the patch strings?

jcansdale: I was wondering about what exactly to validate in this case. The main thing I wanted to validate was that the flag was making its way to indent heuristic code. By checking when we expect the flag to make a difference and when we don't expect it to make a difference, I hope that is pretty much covered.

I've added a sanity check in 2f583cb to make sure that the same lines are being added or removed. I'm trusting the algorithm to add or remove them in the correct order. 🙂

PR #1654, opened 2019-01-14
Use IRepository instance instead of Repository concrete class

nimbusparis:


ethomson:
Unfortunately, `IRepository` doesn't have the [`Handle` field](https://github.com/libgit2/libgit2sharp/pull/1654/files#diff-0281073ed72bd116db776aae26783b68R40).

bording:
Given that `IRepository` doesn't have the required members to make this change, I'm going to go ahead and close the PR for now.

@nimbusparis Feel free to reply about why you wanted to make this change, and maybe we can come up with an alternative.

PR #1655, opened 2019-01-14
Use IRepository instance instead of concrete class

nimbusparis:


bording:
@nimbusparis This PR has similar problems as your other PR, #1654.

I'm going to go ahead and close it for now, but as I mentioned in the other one, let us know what you were trying to achieve and maybe we can come up with an alternative.

PR #1665, opened 2019-02-20
Fix typos in 'Optimizing unit testing'

v-wilock:
"Insure" relates to having insurance. "ensure" means "to make sure of"
"One-thousand" shouldn't be hyphenated. It should only be hyphenated in the case of a compound modifier on the word "thousand" as in "forty-five thousand".


bording:
Thanks for the corrections!

PR #1669, opened 2019-03-11
FileHistory multiple parent commits bug

matlodej:
`IQueryableCommitLog.QueryBy(string path)` does not follow the history of the file if the file was created on another branch and then merged back to source branch using a 3-way merge.

The correct behavior can be confirmed by following the included test and performing `git log <path>` manually, which yields different results than the `QueryBy` method.

The problem is resolved by ensuring all parent commits of a merge commit are observed before exiting the loop if a merge commit is encountered.


bording:
Hi @matlodej I was doing some cleanup and got rid of the old `vNext` branch, which caused your PR to be closed. Can you update your PR to be against `master` instead?

PR #1673, opened 2019-04-20
Add support for Alpine 3.9

bording:
LibGit2Sharp.NativeBinaries 2.0.278 adds an Alpine 3.9 binary, which is needed because it includes OpenSSL 1.1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@
 
 language: csharp
 dist: trusty
-dotnet: 2.1.401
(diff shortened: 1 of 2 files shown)


bording:
Travis appears to still be broken even after applying the suggested fixes from the support forum. The linux tests are running fine on Azure Pipelines, so I'm going to go ahead and merge this and worry about Travis later.

Maybe it's time to just remove it instead of trying to fix it.

PR #1674, opened 2019-04-21
 Housekeeping and cleanup updates

bording:
This is primarily a bit of housekeeping and cleanup:

- Testing and build tool package updates
- Ensuring all tests can pass in the VS test runner by disabling shadow copying
- No need to disable signing since non-Windows platforms can sign now

It also includes a slight change to the version string introduced because of https://github.com/AArnott/Nerdbank.GitVersioning/pull/269

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,7 +1,0 @@
-<Project>
-
-  <PropertyGroup>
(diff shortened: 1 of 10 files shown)


bording:
Once again everything is passing except Travis. Definitely time to remove it!

PR #1679, opened 2019-05-16
update stated platform support

adamralph:
I'm not sure if this is the best way of stating it, but it does bring the statements up to date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,8 @@
 
 ## Prerequisites
 
- - **Windows:** .NET 4.6.1+
- - **Linux/Mac OS X:** Mono 5.4+
+ - **Windows:** .NET 4.6.1+ or any platform supporting .NET Standard 2.0+
+ - **Linux/Mac OS X:** Mono 5.4+ or any platform supporting .NET Standard 2.0+
 
 ## Online resources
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,8 +21,8 @@
 
 ## Prerequisites
 
- - **Windows:** .NET 4.6.1+
- - **Linux/Mac OS X:** Mono 5.4+
+ - **Windows:** .NET 4.6.1+ or any platform supporting .NET Standard 2.0+
```
bording: It actually supports .NET Framework 4.6, so that should also be changed.

bording:
@adamralph This gets a bit tricky due to the native binary requirement. For example even though it has a `netstandard2.0` assembly, actual platform support depends on what native binaries are available.

adamralph:
@bording

> It actually supports .NET Framework 4.6, so that should also be changed.

👍 I've made that change.

> This gets a bit tricky due to the native binary requirement. For example even though it has a `netstandard2.0` assembly, actual platform support depends on what native binaries are available.

True, but I'm afraid I have little knowledge about that all that. I think this PR is a step in the right direction, but please feel free to do what you want with it, including closing it if you'd rather handle this later and all at once.

PR #1686, opened 2019-06-11
Add ARM64 Support & CI

qmfrederik:
libgit2/libgit2sharp.nativebinaries#91 added support for building the native libraries for ARM64.

This PR adds support for building and testing libgit2sharp on ARM64. It uses Drone CI to build & test on native ARM64 hardware. It uses netcoreapp3.0 to run the tests.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,15 @@
+kind: pipeline
+name: default
+
+platform:
+  os: linux
+  arch: arm64
+
+steps:
+- name: build
+  image: mcr.microsoft.com/dotnet/core/sdk:3.0-bionic-arm64v8
+  commands:
+  - date -u
+  - uname -a
+  - env | sort
+  - ./buildandtest.sh 'LEAKS_IDENTIFYING'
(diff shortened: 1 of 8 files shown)


qmfrederik:
@bording I got the build more or less working on ARM64 (1 unit test failing) so I thought I'd put out this PR and see if there's any way we can update libgit2sharp and include ARM64 support?

Of particular interest is the dependency on NerdBank.GitVersioning & SourceLink, both which use libgit2sharp and will fail on ARM64 hardware (since they use a version of libgit2sharp which doesn't have ARM64 support).

Is there any way we could 'side-load' an arm64-version of libgit2 that these tasks can use?

bording:
Lots of things have changed in the repo since this PR was first opened. At this point LibGit2Sharp does have full ARM support.

The one thing we don't have yet is a way to run tests for ARM, but with GitHub Actions saying that ARM runners are coming soon, at this point I'd rather just wait for that vs. introducing a second CI system for it.

Issue #1429, opened 2017-03-23
Question: Getting http.proxy to work


pushad:
Hey guys,

I've got a repo set up at bitbucket. This repository is initialized locally with a reference to the http link provided from the site.

I am using LibGit2Sharp 0.23.1, and I am using the following to fetch from the remote server:
```
private bool fetch()
        {
            string logMessage = "";
            using (var repo = new Repository(settings.directory))
            {
                Remote remote = repo.Network.Remotes["origin"];
                IEnumerable<string> refSpecs = remote.FetchRefSpecs.Select(x => x.Specification);
                FetchOptions options = new FetchOptions();

                options.CredentialsProvider = getCredentials();
                Commands.Fetch(repo, remote.Name, refSpecs, options, logMessage);
            }
            Logger.debug(logMessage);
            return true;
        }
```

Inside config file of git I have the following added:
```
[http]
	proxy = http://<IP>:8888
```

I am using wireshark to sniff my outgoing packets and can confirm that using command-line git fetch, it is indeed using the configured proxy. Removing this setting from the config also confirms that it is not using the proxy. Also at my proxy server I can confirm that there is a connection going to bitbucket when fetching.

However when I use my code, which points to the same location where the repository is sitting at, it does not use the proxy at all. I don't see any packets going to the proxy server in my Wireshark instance, nor in the logs of my proxy server. It is indeed fetching, just directly to the server.

I've googled quite a bit, I've tried setting up environment proxy variable, setting up proxy in Internet Explorer, setting up proxy on remote only. None of them work with my client.

I'm guessing I've got something misconfigured, just not sure what. 

Hopefully you guys can help me out!

bording:
@pushad Taking a look at the code, I think there might be a problem here.

It looks like [`GitProxyOptions`](https://github.com/libgit2/libgit2sharp/blob/2a8b1f023d1ac1660f790f02b9a3f465c02d2b3e/LibGit2Sharp/Core/GitProxyOptions.cs#L14) were added in 88cf9a7b8f1732ae8a2953816895989bb2ae9403, and they are being used in `Commands.Fetch`, as you can see [here](https://github.com/libgit2/libgit2sharp/blob/2a8b1f023d1ac1660f790f02b9a3f465c02d2b3e/LibGit2Sharp/Commands/Fetch.cs#L74).

[...]

PR #1689, opened 2019-06-17
Setting Git proxy type auto where GitProxyOptions 

KashmiraDeshpande:
Setting Git proxy type auto where GitProxyOptions version is set to 1 at object initialization

fix for https://github.com/libgit2/libgit2sharp/issues/1429


LibGit2Sharp/Core/GitProxyOptions.cs
```diff
@@ -14,7 +14,7 @@ internal enum GitProxyType
     internal struct GitProxyOptions
     {
         public uint Version;
-        public GitProxyType Type;
+        public GitProxyType Type= GitProxyType.Auto;
```
lindexi: The code without format

PR #1708, opened 2019-08-21
Proxy options initializing #1429

jwonagel:
Added Factory for initializing Proxyoptions with GitProxyType Auto.
From issue #1429


reegeek:
hi,

When this PR will be merged ?
Regards,

seshonaar:
Please note that when the port is missing in the proxy URL, libgit2 will crash (default_port is not initialized).

seshonaar:
The issue on the libgit2 side is fixed in master, but not on the maintenance branch (0.28.*), that is used by libgit2sharp

jwonagel:
Proxy can be configured in PushOptions, FetchOptionsBase and RemoteOptions. The behavior does not change.

florenzen:
I would really like to see this Pull Request merged and released. We are using libgit2sharp in an internal tool and have absolutely have to use a proxy. Therefore, I applied the PR branch on the v0.26.2 tag in our fork and it works perfectly (see https://github.com/enertrag/libgit2sharp/releases/tag/v0.26.2_proxy-options). But I'd rather reference a NuGet package than including our own build of this library.

PR #1728, opened 2019-10-16
Update libgit2 to ef5a385

ethomson:
Although libgit2 hasn't yet started locking down for its next release, it should be 🔜 .  We can start readying our prereleases to match.


ethomson:
@bording, can I request a quick review so that I can ship a prerelease nuget and unblock @AArnott?  Thanks!

AArnott:
Thank you both!

PR #1741, opened 2019-11-16
NativeMethods.ResolveDll: Avoid exceptions

qmfrederik:
On .NET Core 3.0, LibGit2Sharp registers a per-assembly resolver using via `NativeLibrary.SetDllImportResolver`.

The [per-assembly resolver is the first attempt to resolve native library loads initiated by an assembly](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.nativelibrary.setdllimportresolver?view=netcore-3.0#remarks) and takes precedence over other mechanisms, such as `AssemblyLoadContext`.

The current implementation throws an exception if the `{assemblyDirectory}/runtimes` directory does not exist. In that case, the native load mechanism aborts, and `AssemblyLoadContext` never gets a chance to locate the native library.

This PR adds a `Directory.Exists` check to avoid that exception. `ResolveDll` will return `IntPtr.Zero`, and the `AssemblyLoadContext` is invoked next.

### DIFF
--- a/LibGit2Sharp/Core/NativeMethods.cs
+++ b/LibGit2Sharp/Core/NativeMethods.cs
@@ -156,12 +156,17 @@ namespace LibGit2Sharp.Core
                     // The <rid> ends with the processor architecture. e.g. fedora-x64.
                     string assemblyDirectory = Path.GetDirectoryName(typeof(NativeMethods).Assembly.Location);
                     string processorArchitecture = RuntimeInformation.ProcessArchitecture.ToString().ToLowerInvariant();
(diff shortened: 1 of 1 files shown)


bording:
This is now on nuget.org as 0.27.0-preview-0034.

qmfrederik:
@bording Awesome, thanks!

robertcoltheart:
Can this be either a) implemented in the 0.26 maintenance branch and released to Nuget as `0.26.3`, or b) release `0.27.0` soon? We're struggling with this over here: https://github.com/GitTools/GitVersion/issues/1852

bording:
Given the other work going on to try to improve the overall native binary scenario, I don't think we're at a point where we'd want to do another non-preview release yet.

PR #1759, opened 2020-01-30
Fix up compiler warnings

AArnott:
Fix xml doc compiler warnings


AArnott:
The build failures seem to be due to unstable tests.

PR #1761, opened 2020-01-31
Use Azure Pipelines exclusively (i.e. drop Travis, AppVeyor)

### LINKED ISSUE #1760: Evolve CI builds
I propose to apply some of the pattern from https://github.com/aarnott/library.template to this repo. This will give you Tests recognition within Azure Pipelines. Test results will be captured and represented as well as code coverage.

While I'm at it, I could remove AppVeyor and travis unless it's providing you with some value that can't be obtained from Azure Pipelines.

Thoughts?

AArnott:
Here is a [sample github action result](https://github.com/AArnott/libgit2sharp/actions/runs/33210439). To see this on your *own* Checks tab, I think you'll have to just accept this PR. 
After completing, you'll also want to go delete the hooks from this repo to Travis and AppVeyor so they stop kicking off on pushes or PRs to this repo.

Here is a [sample code coverage report](https://codecov.io/gh/AArnott/libgit2sharp/tree/128b4efb9ea5491f2868a1aec644aa67fd7ba4da/LibGit2Sharp) if we could reactivate it. That will require resolving the problem of the shadow copied assembly failing signing checks.

I don't actually deactivate Azure Pipelines with this change yet. Rather, I enhance the Azure Pipelines integration. Azure Pipelines has a built-in test and test coverage display whereas GitHub Actions does not. We should probably decide which to keep as part of this PR and disable the other.

Closes #1760


bording:
Can you rebase this get rid of the Pinning and fixCompilerWarnings commits?

LibGit2Sharp/LibGit2Sharp.csproj
```diff
@@ -11,7 +11,8 @@
     <Authors>LibGit2Sharp contributors</Authors>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
-    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
+    <IncludeSymbols>true</IncludeSymbols>
```
bording: Let's leave this out for now. I much prefer keeping PDBs in the main package, and this isn't related to changing the CI.

LibGit2Sharp.Tests/StatusFixture.cs
```diff
@@ -640,7 +640,7 @@ public void CanIncludeStatusOfUnalteredFiles()
                 RepositoryStatus status = repo.RetrieveStatus(new StatusOptions() { IncludeUnaltered = true });
 
                 Assert.Equal(unalteredPaths.Length, status.Unaltered.Count());
-                Assert.Equal(unalteredPaths, status.Unaltered.OrderBy(s => s.FilePath).Select(s => s.FilePath).ToArray());
+                Assert.Equal(unalteredPaths, status.Unaltered.OrderBy(s => s.FilePath, StringComparer.InvariantCultureIgnoreCase).Select(s => s.FilePath).ToArray());
```
bording: Let's keep this out of this PR since it's not related to changing the CI.

I'm also thinking that `Ordinal` makes more sense here.

AArnott: I assume you mean OrdinalIgnoreCase. That's fine. But I think the effectiveness of this change proves it's the same across platform.
it is related, because this test failed on Ubuntu until I fixed it.

LibGit2Sharp/LibGit2Sharp.csproj
AArnott: Ok

LibGit2Sharp.Tests/StatusFixture.cs
bording: Actually no, why would we want to ignore case?

AArnott: Because that is required to meet the expected array order in the test as written.

bording: Ok, I took a closer look at the test, and now I can see what's going on, and why it's okay for ignore case.

However, this test hasn't been failing previously AFAIK, so what makes this a problem now that it needs to be part of this PR?

AArnott: It only failed on Ubuntu, which is 18.04 IIRC. Before that you were running on 16.04. My guess is that changes how the file system is laid out such that it produces a different file ordering.

AArnott: Or maybe the default comparer used when not specified somehow has changed. We're still running on netcoreapp2.1 as before, so I can't explain it.

AArnott: I took this commit out of my last push just to see if it fails again, since you didn't like it as part of this.

AArnott: Ya, see, [it failed again](https://github.com/AArnott/libgit2sharp/runs/418610738?check_suite_focus=true):

```
[xUnit.net 00:00:42.09]     LibGit2Sharp.Tests.StatusFixture.CanIncludeStatusOfUnalteredFiles [FAIL]
[xUnit.net 00:00:42.09]       Assert.Equal() Failure
[xUnit.net 00:00:42.09]       Expected: String[] ["1.txt", "1/branch_file.txt", "branch_file.txt", "new.txt", "README"]
[xUnit.net 00:00:42.09]       Actual:   String[] ["1.txt", "1/branch_file.txt", "README", "branch_file.txt", "new.txt"]
[xUnit.net 00:00:42.09]       Stack Trace:
[xUnit.net 00:00:42.09]         /home/runner/work/libgit2sharp/libgit2sharp/LibGit2Sharp.Tests/StatusFixture.cs(629,0): at LibGit2Sharp.Tests.StatusFixture.CanIncludeStatusOfUnalteredFiles()
```

So I'll push the commit back up

bording: Weird, guess we'll keep it in then!

AArnott:
After completing, you'll also want to go delete the hooks from this repo to Travis and AppVeyor so they stop kicking off on pushes or PRs to this repo.

BTW running `dotnet test` on the Azure Pipelines agents anyway will run the tests against mono. But some of the tests failed, so I disabled that. It wasn't something you were already covering before anyway. But if you 'support' this library on mono, I suggest for a subsequent PR we explore those failures and see if we can fix them.

Do you know whether you want to go the Azure Pipelines route or the GitHub Actions route (or both)? You can see how Azure Pipelines now recognizes tests. If you enable analytics for your AZP account, you'll even see which tests are unstable, etc. It's pretty sweet. But GitHub Actions show logs right within GitHub which is also nice. Decisions, decisions. 🤷‍♂

AArnott:
After investigating several test failures in the last 24 hours (usually due to #1764) where both GitHub Actions and Azure Pipelines was in operation, I can heartily suggest we go with Azure Pipelines. Just *searching* for which test failed in the GitHub Actions log is super painful. Contrast that with Azure Pipelines (now with this PR) integrated test results and analytics (when you turn that on) is a supreme experience.

bording:
@AArnott That sounds good to me. Let's drop the GitHub Actions stuff in this PR then.

AArnott:
Don't merge yet. The net46 tests aren't running and I figured out how to get code coverage to work.

AArnott:
Ok, we're reading to merge this. Check this out:

Individual test results:

image

And integrated code coverage:

image

You can also see [an even nicer coverage report](https://codecov.io/gh/AArnott/libgit2sharp/branch/merge) if you'll visit https://codecov.io/gh/libgit2sharp/libgit2sharp/ and just click the 'turn on' button.

AArnott:
Oh, and if you enable the codecov.io integration, your PRs will also get a nice comment [like this](https://github.com/microsoft/vs-streamjsonrpc/pull/407#issuecomment-580858731) so you know what impact the PR has on your code coverage.

bording:
@AArnott Thanks for taking the time to do this, but I think I'm going to need this to be split into a few steps. There are several things in here that seem like unrelated changes that are all done in a single commit with no real explanation.

Also, there are a ton of new scripts that weren't in here before. Why do we suddenly need all of these if we already had Azure Pipelines CI working and running tests? That's a lot to take on and be responsible for maintaining.

AArnott:
Many of these new files come in from my https://github.com/aarnott/library.template template repo, where I refined the art of writing build authoring that runs on all three operating systems, both locally on dev boxes and (minimal) code that only works on Azure Pipelines itself, which makes maintaining and testing build authoring as easy as possible.

This PR brings up your Azure Pipelines YML to cover everything (AFAIK) that AppVeyor and Travis were giving you, drops those other CIs, and yes, it beefs up your Azure Pipelines a bit more. I could add comments to many places in this PR to explain why it's all useful/necessary. I'd rather do that than artificially break it up into smaller chunks.

AArnott:
@bording will you accept a bunch of PR comments explaining what everything does, and then you can decide whether to reject certain parts and I'll remove those?

bording:
> @bording will you accept a bunch of PR comments explaining what everything does, and then you can decide whether to reject certain parts and I'll remove those?

That sounds like a okay way to move forward, yes.

Directory.Build.props
```diff
@@ -4,7 +4,11 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <OutputPath>$(MSBuildThisFileDirectory)bin\$(MSBuildProjectName)\$(Configuration)\</OutputPath>
     <BaseIntermediateOutputPath>$(MSBuildThisFileDirectory)obj\$(MSBuildProjectName)\</BaseIntermediateOutputPath>
+    <PackageOutputPath>$(MSBuildThisFileDirectory)bin\Packages\$(Configuration)\</PackageOutputPath>
     <DefineConstants Condition=" '$(ExtraDefine)' != '' ">$(DefineConstants);$(ExtraDefine)</DefineConstants>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" PrivateAssets="All" />
```
AArnott: This allows targeting .NET Framework on non-Windows platforms.

Directory.Build.props
```diff
@@ -4,7 +4,11 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <OutputPath>$(MSBuildThisFileDirectory)bin\$(MSBuildProjectName)\$(Configuration)\</OutputPath>
     <BaseIntermediateOutputPath>$(MSBuildThisFileDirectory)obj\$(MSBuildProjectName)\</BaseIntermediateOutputPath>
+    <PackageOutputPath>$(MSBuildThisFileDirectory)bin\Packages\$(Configuration)\</PackageOutputPath>
```
AArnott: This collects all the artifacts (all 1 of them) into a unique directory that can be directly collected as an artifact in Azure Pipelines so that deployment to nuget.org is easier, and even automated.

Directory.Build.targets
```diff
@@ -0,0 +1,8 @@
+<Project>
+  <PropertyGroup Condition=" '$(IsTestProject)' == 'true' ">
+    <CoverletOutputFormat>cobertura</CoverletOutputFormat>
+    <Exclude>[xunit.*]*</Exclude>
+    <!-- Ensure we preserve each coverlet output file per target framework: https://github.com/tonerdo/coverlet/issues/177 -->
+    <CoverletOutput>$(OutputPath)/</CoverletOutput>
+  </PropertyGroup>
```
AArnott: This whole file enables code coverage collection when `/p:CollectCoverage=true` is specified, as is now done in the Azure Pipelines .yml file.

LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
```diff
@@ -11,7 +11,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.0.1" />
+    <PackageReference Include="coverlet.msbuild" Version="2.7.0" />
```
AArnott: This is required for code coverage.

LibGit2Sharp.Tests/StatusFixture.cs
```diff
@@ -640,7 +640,7 @@ public void CanIncludeStatusOfUnalteredFiles()
                 RepositoryStatus status = repo.RetrieveStatus(new StatusOptions() { IncludeUnaltered = true });
 
                 Assert.Equal(unalteredPaths.Length, status.Unaltered.Count());
-                Assert.Equal(unalteredPaths, status.Unaltered.OrderBy(s => s.FilePath).Select(s => s.FilePath).ToArray());
+                Assert.Equal(unalteredPaths, status.Unaltered.OrderBy(s => s.FilePath, StringComparer.OrdinalIgnoreCase).Select(s => s.FilePath).ToArray());
```
AArnott: This fixes a test that (oddly) fails only on Azure Pipelines' Linux agent.

LibGit2Sharp.Tests/desktop/ShadowCopyFixture.cs
```diff
@@ -11,6 +11,7 @@ namespace LibGit2Sharp.Tests
     public class ShadowCopyFixture : BaseFixture
     {
         [Fact]
+        [Trait("TestCategory", "FailsWhileInstrumented")]
```
AArnott: This test sets up an appdomain and loads libgit2sharp. When code coverage is turned on, it invalidates the signature of the strong-name signed assembly and then .NET Framework won't load it. So skip the test during instrumented test runs. We run this test by itself in the very next step on Azure Pipelines to make sure it runs.

LibGit2Sharp/LibGit2Sharp.csproj
```diff
@@ -14,10 +14,13 @@
     <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\libgit2sharp.snk</AssemblyOriginatorKeyFile>
+    <PackageIcon>square-logo.png</PackageIcon>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
```
AArnott: The change to package license and icon is required by the newer .NET SDK we've upgraded to.

azure-pipelines.yml
```diff
@@ -1,34 +1,19 @@
...
+    exclude:
+    - doc/
+    - '*.md'
+    - .vscode/
 
 variables:
-  solution: '**/*.sln'
-  buildPlatform: 'Any CPU'
-  buildConfiguration: 'Release'
+  TreatWarningsAsErrors: true
```
AArnott: `TreatWarningsAsErrors` to keep your build clean. You had a few warnings before I fixed them in #1221, and this helps slam the door against such engineering debt in the future.

azure-pipelines.yml
```diff
@@ -1,34 +1,19 @@
...
+  DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
```
AArnott: This makes your builds just a little faster by skipping a ~20 seconds unpack step when dotnet.exe is used for the first time. That 20 seconds is great for dev boxes where it will pay for itself over time, but on a CI machine, it's a waste.

azure-pipelines.yml
```diff
@@ -1,34 +1,19 @@
...
+  BuildConfiguration: Release
+  NUGET_PACKAGES: $(Agent.TempDirectory)/.nuget/packages
```
AArnott: Divert restored packages to a unique and clean cache location. This ensures both that you haven't hard-coded paths to `$(UserProfile)` to get packages and (more importantly) that you don't assume packages are present without references to them and get away with it just because the agents already have those packages restored for some unrelated reason. In short, it keeps your build authoring clean.

azure-pipelines/artifacts/_all.ps1
```diff
@@ -0,0 +1,50 @@
+# This script returns all the artifacts that should be collected after a build.
```
AArnott: As documented, this script executes all the other scripts in this directory to aggregate the data on artifacts to be uploaded.
Overall, these artifacts scripts make it easier to pick and choose which files to collect, gives an opportunity to create/process files before collection, and allows local dev box testing of all this behavior instead of having to run it through Azure Pipelines just to test a yml script.

azure-pipelines/artifacts/_pipelines.ps1
```diff
@@ -0,0 +1,10 @@
+# This script translates all the artifacts described by _all.ps1
+# into commands that instruct Azure Pipelines to actually collect those artifacts.
```
AArnott: As documented, this is the Azure-Pipelines adapter for the otherwise CI-agnostic artifact scripts. It assigns a special suffix to each artifact based on the OS of the agent so that you can collect artifacts from win, osx, linux without them stomping on each other.

azure-pipelines/artifacts/_stage_all.ps1
```diff
@@ -0,0 +1,59 @@
+# This script links all the artifacts described by _all.ps1
+# into a staging directory, reading for uploading to a cloud build artifact store.
+# It returns a sequence of objects with Name and Path properties.
```
AArnott: This script copies all artifacts into a dedicated directory for each one in preparation for capture by whatever CI system you use.

azure-pipelines/artifacts/build_logs.ps1
```diff
@@ -0,0 +1,12 @@
+if ($env:BUILD_ARTIFACTSTAGINGDIRECTORY) {
```
AArnott: This collects all the build logs for analysis when something goes wrong in CI.

azure-pipelines/artifacts/coverageResults.ps1
```diff
@@ -0,0 +1,22 @@
+$RepoRoot = [System.IO.Path]::GetFullPath("$PSScriptRoot\..\..")
```
AArnott: This collects the coverage results, which are later merged from all 3 agents into one report and published to Azure Pipelines' built in coverage viewer.

azure-pipelines/artifacts/deployables.ps1
```diff
@@ -0,0 +1,13 @@
+$RepoRoot = [System.IO.Path]::GetFullPath("$PSScriptRoot\..\..")
```
AArnott: This collects everything you might deploy. For your repo, that's just the nupkg files.

azure-pipelines/artifacts/projectAssetsJson.ps1
```diff
@@ -0,0 +1,9 @@
+$ObjRoot = [System.IO.Path]::GetFullPath("$PSScriptRoot\..\..\obj")
```
AArnott: This collects all your project.assets.json files, which can be useful for diagnosing failures that might be unique to your CI machine.

azure-pipelines/dotnet.yml
```diff
@@ -0,0 +1,58 @@
+steps:
+
+- script: dotnet build --no-restore -c $(BuildConfiguration) /v:m /bl:"$(Build.ArtifactStagingDirectory)/build_logs/build.binlog"
```
AArnott: This and every other build step we use `/bl` to collect the most exhaustive build log possible for diagnosing failures.
We skip restore since that was done in a prior step and don't want to spend time restoring again.

azure-pipelines/dotnet.yml
```diff
@@ -0,0 +1,58 @@
...
+  displayName: dotnet build
+
+- script: dotnet pack --no-build -c $(BuildConfiguration) /v:m /bl:"$(Build.ArtifactStagingDirectory)/build_logs/pack.binlog"
+  displayName: dotnet pack
+
+- task: DotNetCoreCLI@2
+  displayName: dotnet test -f net46 (w/ coverage)
```
AArnott: Testing on .NET Framework is split in two chunks: w/ coverage and w/o coverage. Tests aren't duplicated across these two chunks, but they have to be separated because there's one test that fails when coverage is enabled.

azure-pipelines/dotnet.yml
```diff
@@ -0,0 +1,58 @@
...
+- task: DotNetCoreCLI@2
+  displayName: dotnet test -f net46 (w/o coverage)
+  inputs:
+    command: test
+    arguments: --no-build -c $(BuildConfiguration) -f net46 --filter "TestCategory!=FailsInCloudTest & TestCategory=FailsWhileInstrumented" -v n
+    testRunTitle: net46-$(Agent.JobName)-nocoverage
+  condition: and(succeeded(), eq(variables['Agent.OS'], 'Windows_NT'))
+
+- task: DotNetCoreCLI@2
+  displayName: dotnet test -f netcoreapp2.1
```
AArnott: By using the `DotNetCoreCLI` task for these tests, AzP will automatically collect test results and report per-test results and analytics about test stability.

azure-pipelines/dotnet.yml
```diff
@@ -0,0 +1,58 @@
...
+- task: PowerShell@2
+  inputs:
+    filePath: azure-pipelines/variables/_pipelines.ps1
+    failOnStderr: true
+  displayName: Update pipeline variables based on build outputs
+  condition: succeededOrFailed()
+
+- task: PowerShell@2
+  inputs:
+    filePath: azure-pipelines/artifacts/_pipelines.ps1
```
AArnott: This script will process, prepare and collect all Azure Pipelines artifacts in one step.

azure-pipelines/dotnet.yml
```diff
@@ -0,0 +1,58 @@
...
+- task: PublishSymbols@2
+  inputs:
+    SymbolsFolder: $(Build.ArtifactStagingDirectory)/symbols-Windows
+    SearchPattern: '**/*.pdb'
+    IndexSources: false
+    SymbolServerType: TeamServices
+  displayName: Publish symbols to symbol server
+  condition: and(succeeded(), ne(variables['Build.Reason'], 'PullRequest'), eq(variables['Agent.OS'], 'Windows_NT'))
+
+- bash: bash <(curl -s https://codecov.io/bash)
```
AArnott: This publishes per-agent coverage reports to codecov.io. This gives you:
1. a great view of your coverage on any branch, commit and PR.
2. An automated PR comment on each PR explaining the impact to code coverage the change has.

azure-pipelines/install-dependencies.yml
```diff
@@ -0,0 +1,15 @@
+parameters:
+  initArgs:
+
+steps:
+
+- powershell: |
+    .\init.ps1 -AccessToken '$(System.AccessToken)' ${{ parameters['initArgs'] }} -UpgradePrerequisites
```
AArnott: This script ensure that the agent (or your own dev box if you run it directly) has all the dependencies necessary. Most importantly being the .NET Core SDK and runtimes.

azure-pipelines/publish-codecoverage.yml
```diff
@@ -0,0 +1,31 @@
...
+  continueOnError: true
+- download: current
+  artifact: coverageResults-Linux
+  displayName: Download Linux code coverage results
+  continueOnError: true
+- download: current
+  artifact: coverageResults-macOS
+  displayName: Download macOS code coverage results
+  continueOnError: true
+- powershell: |
```
AArnott: This is the script that merges the coverage reports from all 3 agents into one and publishes them to Azure Pipelines for its integrate coverage reports.

azure-pipelines/publish-deployables.yml
```diff
@@ -0,0 +1,14 @@
...
+  displayName: Download deployables
+  artifact: deployables-Windows
+
+- task: NuGetCommand@2
+  displayName: Push packages to CI feed
+  inputs:
+    command: push
+    packagesToPush: $(Pipeline.Workspace)/deployables-Windows/*.nupkg
+    nuGetFeedType: internal
+    publishVstsFeed: $(ci_feed)
```
AArnott: If you set `ci_feed` to the name of an Azure Artifacts feed, this will automatically light up and push your nupkg there.
If you have a public project on Azure Pipelines where this build occurs, this feed can be public as well, giving folks a place to get the latest CI builds of your library. This is a boon for folks who don't want to wait for your (I think quarterly?) scheduled releases to nuget.org

azure-pipelines/variables/DotNetSdkVersion.ps1
```diff
@@ -0,0 +1,2 @@
+$globalJson = Get-Content -Path "$PSScriptRoot\..\..\global.json" | ConvertFrom-Json
+$globalJson.sdk.version
```
AArnott: This script is used by `init.ps1` to install whatever .NET Core SDK happens to be required per your global.json file.

init.cmd
```diff
@@ -0,0 +1,4 @@
+@echo off
+SETLOCAL
+set PS1UnderCmd=1
+powershell.exe -NoProfile -NoLogo -ExecutionPolicy bypass -Command "try { & '%~dpn0.ps1' %*; exit $LASTEXITCODE } catch { write-host $_; exit 1 }"
```
AArnott: For folks who prefer cmd.exe's over powershell for their CLI, this is a convenient stub that invokes init.ps1 in a natural way.

init.ps1
```diff
@@ -0,0 +1,67 @@
+<#
+.SYNOPSIS
+Installs dependencies required to build and test the projects in this repository.
```
AArnott: After running this script, *any* machine should be ready to build and run your tests.

init.ps1
```diff
@@ -0,0 +1,67 @@
...
+.DESCRIPTION
+This MAY not require elevation, as the SDK and runtimes are installed to a per-user location,
+unless the `-InstallLocality` switch is specified directing to a per-repo or per-machine location.
+See detailed help on that switch for more information.
+.PARAMETER InstallLocality
```
AArnott: Do you trust OSS projects that you can't build without giving them elevated admin permissions to install stuff? I don't. This script defaults to *not* requiring elevation by installing dependencies in non-privileged locations. But you *can* opt into installing at per-machine locations for greater convenience. See docs in this script for more info.

AArnott:
So what do you think of my own code review comments above, @bording?

bording:
I haven't had time to look at them yet, but I hope to soon. Thanks for taking the time to do that.

LibGit2Sharp/LibGit2Sharp.csproj
bording: Instead of using a license expression, I'd prefer this to embed the actual license file from the repo.

azure-pipelines.yml
bording: Instead of putting this here, I think it makes sense to add `<TreatWarningsAsErrors>true</TreatWarningsAsErrors>` to Directory.Build.props.

bording:
@AArnott Looking over the stuff here, I do have some general thoughts/questions.

It's unfortunate that adding the code coverage stuff requires all the extra complexity of multiple steps and scripts, etc. I'm not really used to using code coverage tools, so it makes me wonder if it's worth the cost.

We currently have buildandtest.cmd / buildandtest.sh that were previously used by the CI scripts and could also be used locally. These are no longer used, right? How would someone locally run a build that is the same as what CI would do then?

LibGit2Sharp/LibGit2Sharp.csproj
AArnott: That is certainly an option, and the right choice if you have a non-standard license. Nuget documentation encourages you to use the license expression if it's a conforming license. Do you want me to embed the license anyway?

azure-pipelines.yml
AArnott: The problem with that is then you can't even build locally when there are warnings. That slows down your dev inner loop. I greatly prefer for warnings to be allowed during the inner loop and only be accountable to fix all warnings before a PR is accepted.

LibGit2Sharp/LibGit2Sharp.csproj
bording: Since we're already including the license file in the package anyway, I'd prefer to link to it.

Overall I'm not really a fan of how the license expressions work and how nuget.org then displays a generic license template vs. your actual license text.

AArnott: OK. I'll link to the embedded license.

AArnott:
> It's unfortunate that adding the code coverage stuff requires all the extra complexity of multiple steps and scripts, etc. I'm not really used to using code coverage tools, so it makes me wonder if it's worth the cost.

@bording The code coverage is only half of it. The other half of the extra steps is due to simply wanting individual test run reports per-framework and per agent. 

> We currently have buildandtest.cmd / buildandtest.sh that were previously used by the CI scripts and could also be used locally. These are no longer used, right?

They're not used in CI any more, that's correct. They *could* be, but then the test results wouldn't be included nicely by Azure Pipelines' test reports. But end users can certainly use them if they're useful. Or we can remove them. 

> How would someone locally run a build that is the same as what CI would do then?

Build or test with `dotnet build` or `dotnet test` does effectively the same thing as CI (in a simpler way, since we don't need CI reporting when local tests run). 

CI runs on all 3 agents of course, and a 4th agent to run with `/p:ExtraDefine=LEAKS_IDENTIFYING`. Even the original 2 buildandtest scripts didn't specify that without an extra argument passed to the scripts. 

So we can decide whether to keep the scripts (which IMO don't add any value) or remove them.

bording:
> @bording The code coverage is only half of it. The other half of the extra steps is due to simply wanting individual test run reports per-framework and per agent.

True, but there would definitely be fewer changes and steps if the coverage stuff wasn't included in this.

However, after thinking about it for a bit, I think it's not worth the effort to get rid of it since you've already got all of it working.

> So we can decide whether to keep the scripts (which IMO don't add any value) or remove them.

Let's go ahead and remove them as part of this PR and see how it goes.

Once you remove those, I think I'm ready to approve and merge this, though I reserve the right to be able to ask you for help with any of these scripts if something goes wrong in the future! :smile:

PR #1762, opened 2020-01-31
Improve TruncateMilliseconds test helper

bording:
Looking at the flaky test failures, my theory is that the previous implementation of `TruncateMilliseconds` is the source of the problem. Let's see if this helps.


bording:
Well, it appears this isn't the source of the problem. This change is still a better way to implement the method, so I'm going to go ahead and merge it.

PR #1772, opened 2020-03-02
Suppress Expect: 100 Continue header

AArnott:
This removes the unused `HttpClient` fields too.


bording:
@AArnott Looks like making the change for the `MemoryStream` has introduced conflicts here. I'll merge this once they are resolved!

AArnott:
Done.

Issue #1778, opened 2020-04-04
Patch diff with ignore white space


Divad-H:
When creating a patch diff using `Diff.Compare` I did not find any option to ignore white space.

I think `CompareOptions` should contain an enum corresponding to the git diff settings:
```
--ignore-all-space
--ignore-space-change
--ignore-space-at-eol
```

Currently there is a `WhitespaceMode` in `SimilarityOptions`. However, this is intended for handling white space while detecting renames and copies.

I intend to submit a pull request for this.

### Version of LibGit2Sharp
4de700fe

darthkurak:
I will bring this up.
Is there any chance to support --ignore-all-space?

Divad-H:
I had also opened a PR for this when I opened the issue: 
https://github.com/libgit2/libgit2sharp/pull/1779

SenorBasilico:
What needs to happen for this pull request to be merged?

PR #1779, opened 2020-04-04
White space diff options #1778

Divad-H:
I was unsure if I should create a separate file for the enum `PatchWhitespaceMode`.


Divad-H:
Address https://github.com/libgit2/libgit2sharp/issues/1778

PR #1795, opened 2020-05-13
Make all Commands methods use IRepository for consistency

asbjornu:
Make all `Commands` methods use `IRepository` for consistency.


asbjornu:
Seems like the following line makes it impossible to use `IRepository` since it does not have a `Handle` property.

https://github.com/libgit2/libgit2sharp/blob/7fc4be5193dbdd08538b4b150332b5a73770e0f6/LibGit2Sharp/Commands/Fetch.cs#L40

Ideas for how to resolve this would be appreciated.

PR #1849, opened 2020-11-11
Bruce mellows/1848

BruceMellows:
Invalid escape in .git/config causes access violation in libgit2


BruceMellows:
I think that this must be fixed in libgit2

PR #1853, opened 2020-11-19
Jwonagel/proxy options initializing

jwonagel:
Made a proxy configurable. Default behavior is the same.


bording:
Thanks for the PR. I've incorporated the sort of things provided here in my comprehensive PR that adds proxy support, #2065.

PR #1895, opened 2021-07-19
Move credential caching to ManagedHttpSmartSubtransport, new cache per instance.

### LINKED ISSUE #1894: Credentials being cached between requests to separate remotes with the same URL
The current implementation of `ManagedHttpSmartSubtransportStream` uses a [static `CredentialCache`](https://github.com/libgit2/libgit2sharp/blob/7c78387a3b5c9483f6aa609aa09b19b78077018a/LibGit2Sharp/Core/ManagedHttpSmartSubtransport.cs#L56). The credentials in this cache are only differentiated by URL and authentication scheme, if you have two separate repositories configured with the same URL, it's possible to authenticate to a repository without providing valid credentials.

### Reproduction steps
_Fill this with the relevant credentials or tokens for your repo, obviously this is terribly insecure for anything real-world_

```csharp
using System;
using System.Net;
using LibGit2Sharp;

namespace Example
{
    class Program
    {
        static void Main(string[] args)
        {
            var uri = "<repository URL>";

            var repo1 = Repository.Clone(uri, "directory1", new CloneOptions
            {
                BranchName = "main",
                CredentialsProvider = CredentialsHandlerThatReturnsValidCredentials
            }); // This will succeed

            var repo2 = Repository.Clone(uri, "directory2", new CloneOptions
            {
                BranchName = "main",
                CredentialsProvider = CredentialHandlerThatReturnsInvalidCredentials
            }); // Invalid credential provider is never called, cached credentials from the first request are used
        }

        public static Credentials CredentialsHandlerThatReturnsValidCredentials(string url, string usernameFromUrl,
            SupportedCredentialTypes types)
        {
            return new SecureUsernamePasswordCredentials
            {
                Username = "<valid username>",
                Password = new NetworkCredential(string.Empty, "<valid password>").SecurePassword
            };
        }

        public static Credentials CredentialHandlerThatReturnsInvalidCredentials(string url, string usernameFromUrl,
            SupportedCredentialTypes types)
        {
            return new SecureUsernamePasswordCredentials
            {
                Username = "invalid username",
                Password = new NetworkCredential(string.Empty, "invalid password").SecurePassword
            };
        }
    }
}
```

### Expected behavior

Second clone should not authenticate, invalid credentials supplied.

### Actual behavior

Authenticates with repository successfully, as there has already been a successful authentication to that repository URL.

`0.27.0-preview-0102`

macOS, .NET 5

dylanlerch:
Fixes #1894

When making multiple requests to the same repository URL with different credentials, the first set of successful credentials would be reused (as the credential cache is static). This can cause unexpected re-use of credentials when the same URL is passed to independent git operations.

This change creates a new HttpClientHandler and CredentialHandler for each ManagedHttpSmartSubtransport to eliminate any sharing of credentials between operations. It's possible that this could be a breaking change for some users; if a consumer is only setting valid credentials for the first operation, subsequent operations will now fail.

This change will result in more calls to the credential provider. If the provider is doing complex authentication logic (potentially calling external services), they will need to be responsible for caching between operations.

### DIFF
--- a/LibGit2Sharp/Core/ManagedHttpSmartSubtransport.cs
+++ b/LibGit2Sharp/Core/ManagedHttpSmartSubtransport.cs
@@ -9,6 +9,31 @@ namespace LibGit2Sharp.Core
 {
     internal class ManagedHttpSmartSubtransport : RpcSmartSubtransport
     {
+#if NETCOREAPP
+        private readonly SocketsHttpHandler httpHandler;
+#else
+        private readonly HttpClientHandler httpHandler;
+#endif
+
+        private readonly CredentialCache credentialCache;
+        private readonly object credentialCacheLock = new object();
+
+        public ManagedHttpSmartSubtransport()
+        {
+#if NETCOREAPP
+            httpHandler = new SocketsHttpHandler();
+            httpHandler.PooledConnectionLifetime = TimeSpan.FromMinutes(5);
+#else
+            httpHandler = new HttpClientHandler();
+            httpHandler.SslProtocols |= SslProtocols.Tls12;
+#endif
+
+            httpHandler.AllowAutoRedirect = false;
+
+            credentialCache = new CredentialCache();
+            httpHandler.Credentials = credentialCache;
+        }
+
         protected override SmartSubtransportStream Action(string url, GitSmartSubtransportAction action)
         {
             string endpointUrl, contentType = null;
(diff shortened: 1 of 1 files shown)


dylanlerch:
Closing this, there are plenty of issues with approach. Any real fix will be completely different and will make sense as a separate PR.

PR #1901, opened 2021-08-13
Update to LibGit2Sharp.NativeBinaries 2.0.313

bording:
The new native binaries package is now compatible with .NET 5 single file publish when building a Windows executable. You will no longer have a native PDB sitting next to your .exe that is required to be there for the .exe to work.


bording:
LibGit2Sharp.NativeBinaries 2.0.313 Has a bad `win-x86` binary. It's actually 64-bit. :cry:

PR #1905, opened 2021-08-22
Update LibGit2 to version 1.1.1

jairbubbles:
- Use package produced by https://github.com/libgit2/libgit2sharp.nativebinaries/pull/114
- Rename some native methods an change return type (see https://github.com/libgit2/libgit2/pull/5117 and https://github.com/libgit2/libgit2/pull/5365)
  -  Thanks @A-Ovchinnikov-mx for poviding the fix! (cherry-picked from https://github.com/mendix/libgit2sharp/commit/f958478f89fa37e80881a63a621b6a56712297ec)
- Update some tests to match the fact that "origin/HEAD" is now added when cloning (see libgit2/libgit2#5651)

### DIFF
--- a/LibGit2Sharp.Tests/CloneFixture.cs
+++ b/LibGit2Sharp.Tests/CloneFixture.cs
@@ -70,7 +70,7 @@ namespace LibGit2Sharp.Tests
                 Assert.NotEqual(originalRepo.Info.Path, clonedRepo.Info.Path);
                 Assert.Equal(originalRepo.Head, clonedRepo.Head);
 
-                Assert.Equal(originalRepo.Branches.Count(), clonedRepo.Branches.Count(b => b.IsRemote));
+                Assert.Equal(originalRepo.Branches.Count(), clonedRepo.Branches.Count(b => b.IsRemote && b.FriendlyName != "origin/HEAD"));
                 Assert.Equal(isCloningAnEmptyRepository ? 0 : 1, clonedRepo.Branches.Count(b => !b.IsRemote));
 
                 Assert.Equal(originalRepo.Tags.Count(), clonedRepo.Tags.Count());
--- a/LibGit2Sharp/Core/NativeMethods.cs
+++ b/LibGit2Sharp/Core/NativeMethods.cs
@@ -227,7 +227,7 @@ namespace LibGit2Sharp.Core
         internal static extern unsafe GitError* git_error_last();
 
         [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern void git_error_set_str(
+        internal static extern int git_error_set_str(
             GitErrorCategory error_class,
             [MarshalAs(UnmanagedType.CustomMarshaler, MarshalCookie = UniqueId.UniqueIdentifier, MarshalTypeRef = typeof(StrictUtf8Marshaler))] string errorString);
 
@@ -252,25 +252,25 @@ namespace LibGit2Sharp.Core
         internal static extern unsafe void git_blame_free(git_blame* blame);
 
         [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern unsafe int git_blob_create_fromdisk(
+        internal static extern unsafe int git_blob_create_from_disk(
             ref GitOid id,
             git_repository* repo,
             [MarshalAs(UnmanagedType.CustomMarshaler, MarshalCookie = UniqueId.UniqueIdentifier, MarshalTypeRef = typeof(StrictFilePathMarshaler))] FilePath path);
 
         [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern unsafe int git_blob_create_fromworkdir(
+        internal static extern unsafe int git_blob_create_from_workdir(
             ref GitOid id,
             git_repository* repo,
             [MarshalAs(UnmanagedType.CustomMarshaler, MarshalCookie = UniqueId.UniqueIdentifier, MarshalTypeRef = typeof(StrictFilePathMarshaler))] FilePath relative_path);
 
         [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern unsafe int git_blob_create_fromstream(
+        internal static extern unsafe int git_blob_create_from_stream(
             out IntPtr stream,
             git_repository* repositoryPtr,
             [MarshalAs(UnmanagedType.CustomMarshaler, MarshalCookie = UniqueId.UniqueIdentifier, MarshalTypeRef = typeof(StrictUtf8Marshaler))] string hintpath);
 
         [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern int git_blob_create_fromstream_commit(
+        internal static extern int git_blob_create_from_stream_commit(
             ref GitOid oid,
             IntPtr stream);
 
@@ -1616,7 +1616,7 @@ namespace LibGit2Sharp.Core
         internal static extern unsafe FilePath git_repository_path(git_repository* repository);
 
         [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern unsafe void git_repository_set_config(
+        internal static extern unsafe int git_repository_set_config(
             git_repository* repository,
             git_config* config);
 
@@ -1628,7 +1628,7 @@ namespace LibGit2Sharp.Core
 
 
         [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern unsafe void git_repository_set_index(
+        internal static extern unsafe int git_repository_set_index(
             git_repository* repository,
             git_index* index);
 
@@ -1710,13 +1710,13 @@ namespace LibGit2Sharp.Core
         internal static extern unsafe int git_revwalk_push(git_revwalk* walker, ref GitOid id);
 
         [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern unsafe void git_revwalk_reset(git_revwalk* walker);
+        internal static extern unsafe int git_revwalk_reset(git_revwalk* walker);
 
         [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern unsafe void git_revwalk_sorting(git_revwalk* walk, CommitSortStrategies sort);
+        internal static extern unsafe int git_revwalk_sorting(git_revwalk* walk, CommitSortStrategies sort);
 
         [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern unsafe void git_revwalk_simplify_first_parent(git_revwalk* walk);
+        internal static extern unsafe int git_revwalk_simplify_first_parent(git_revwalk* walk);
 
         [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
         internal static extern unsafe void git_signature_free(git_signature* signature);
--- a/LibGit2Sharp.Tests/FetchFixture.cs
+++ b/LibGit2Sharp.Tests/FetchFixture.cs
@@ -215,7 +215,7 @@ namespace LibGit2Sharp.Tests
 
             using (var clonedRepo = new Repository(clonedRepoPath))
             {
-                Assert.Equal(5, clonedRepo.Branches.Count(b => b.IsRemote));
+                Assert.Equal(5, clonedRepo.Branches.Count(b => b.IsRemote && b.FriendlyName != "origin/HEAD"));
 
                 // Drop one of the branches in the remote repository
                 using (var sourceRepo = new Repository(source))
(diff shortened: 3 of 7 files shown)


skysb:
Hi, is there an update on this draft or any ETA on when it will be merged to master? It contains an important update to allow git version 1 which would be amazing to have for sparse checkouts, partial clones etc.

jairbubbles:
Hi @skysb! The PR is pretty new so give some time to the maintainer 😅.

It's still draft as it's using as pre-release version of `LibGit2Sharp.NativeBinaries`. 

@bording Do you think it's possible to publish a pre-release version of this package? Also I feel like it would make sense to publish the current as a stable `0.27.0` and `0.28.0-preview` would be for the next version with a recent `libgit2`.

LibGit2Sharp.Tests/CloneFixture.cs
```diff
@@ -70,7 +70,7 @@ private void AssertLocalClone(string url, string path = null, bool isCloningAnEm
                 Assert.NotEqual(originalRepo.Info.Path, clonedRepo.Info.Path);
                 Assert.Equal(originalRepo.Head, clonedRepo.Head);
 
-                Assert.Equal(originalRepo.Branches.Count(), clonedRepo.Branches.Count(b => b.IsRemote));
+                Assert.Equal(originalRepo.Branches.Count(), clonedRepo.Branches.Count(b => b.IsRemote && b.FriendlyName != "origin/HEAD"));
```
bording, line 73: Why were these changes to the tests required?

LibGit2Sharp/Core/NativeMethods.cs
```diff
@@ -227,7 +227,7 @@ private sealed class NativeShutdownObject : CriticalFinalizerObject
         internal static extern unsafe GitError* git_error_last();
 
         [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern void git_error_set_str(
+        internal static extern int git_error_set_str(
```
bording, line 230: I assume the signature changed on these native methods, or was there some other reason for the `void` -> `int` changes?

bording:
> Do you think it's possible to publish a pre-release version of this package?

Yes, I should be able to publish a new preview once this gets merged.

> Also I feel like it would make sense to publish the current as a stable 0.27.0 and 0.28.0-preview would be for the next version with a recent libgit2.

I'm a bit reluctant to do that as this point because the 0.27.0 preview packages include the change over to a managed HTTPS implementation that is definitely a bit experimental, and does seem to be causing some problems. I'd prefer to wait for https://github.com/libgit2/libgit2/pull/5974 to be merged, pull that into the native binaries package, and then revert the managed HTTPs changes.

LibGit2Sharp/Core/NativeMethods.cs
jairbubbles: I took the code from @A-Ovchinnikov-mx. I don't think it changed recently.

LibGit2Sharp.Tests/CloneFixture.cs
jairbubbles: Because of this change on libgit2: https://github.com/libgit2/libgit2/pull/5651

jairbubbles:
> Yes, I should be able to publish a new preview once this gets merged.

But I don't think we can merge with the prerelease package referenced?

bording:
> But I don't think we can merge with the prerelease package referenced?

Why not? The prerelease native package is on nuget.org, so a preview LibGit2Sharp could be released that references it.

LibGit2Sharp/Core/NativeMethods.cs
A-Ovchinnikov-mx: Yep, please see https://github.com/mendix/libgit2sharp/pull/3/files
And https://github.com/libgit2/libgit2/blob/main/docs/changelog.md#v099
I hope we didn't miss anything 😅

LibGit2Sharp.Tests/CloneFixture.cs
A-Ovchinnikov-mx: I'd _probably_ recommend checking the type of Reference instead:
```cs
... (b => b.IsRemote && b.Reference is DirectReference)
```

jairbubbles:
> Why not? The prerelease native package is on nuget.org, so a preview LibGit2Sharp could be released that references it.

Let me undraft then! 😊

bording (approved):

LibGit2Sharp.Tests/CloneFixture.cs
jairbubbles: Thx I didn't know about that. 

I also agree, I've always found weird that it was returning HEAD by default, but it's a breaking change so it should be discussed in another PR.

bording:
0.27.0-preview-0119 is up on nuget.org now.

alfonsarmi:
Thanks for this!

In Windows is working like a charm with partial cloned repos but in Linux I'm having the next error message:

LibGit2Sharp.LibGit2SharpException: 'unsupported extension name extensions.partialclone'

Any clue?

ethomson:
> Hi, is there an update on this draft or any ETA on when it will be merged to master? It contains an important update to allow git version 1 which would be amazing to have for sparse checkouts, partial clones etc.
 
libgit2 doesn't support partial clones or sparse checkouts.  I think that the update that you're referring to allows us to understand the repository format v1, which adds in new optional extensions (like partial clones or sparse checkouts).  The update in libgit2 only allows us to know that we _don't_ support the repository's extensions.

skysb:
> > Hi, is there an update on this draft or any ETA on when it will be merged to master? It contains an important update to allow git version 1 which would be amazing to have for sparse checkouts, partial clones etc.
> 
> libgit2 doesn't support partial clones or sparse checkouts. I think that the update that you're referring to allows us to understand the repository format v1, which adds in new optional extensions (like partial clones or sparse checkouts). The update in libgit2 only allows us to know that we _don't_ support the repository's extensions.

Hi @ethomson , You are right, I didn't explain myself correctly in the post. The idea was to be able to identify v0 from v1 only, as when we tried instantiating a new repository with a partially cloned repository, we got an error saying it wasn't supported.

skysb:
> Thanks for this!
> 
> In Windows is working like a charm with partial cloned repos but in Linux I'm having the next error message:
> 
> LibGit2Sharp.LibGit2SharpException: 'unsupported extension name extensions.partialclone'
> 
> Any clue?

Hi @alfonsarmi ,

I am getting the same issue. When trying to instantiate a partially cloned repository, I get the same error.

@jairbubbles  would this be something related to the native binaries? as in windows it works well but in linux (docker images) it fails for me with the error

`LibGit2Sharp.LibGit2SharpException: 'unsupported extension name extensions.partialclone'`

jairbubbles:
> @jairbubbles would this be something related to the native binaries? as in windows it works well but in linux (docker images) it fails for me with the error

I don't explain why there's a different behavior between Linux and Windows. I quickly looked at the code and there's no `#IFDEF` around it. (https://github.com/libgit2/libgit2/blob/main/src/repository.c#L1669). 

If I understand correctly @ethomson's answer, **we should get the error 'unsupported extension name extensions.partialclone' on both platforms**.

For my use case, I was also able to open a partially cloned repository and access the info I needed (on Windows). Please note that I don't use libgit2sharp to run actions on the repository, it's only used to inspect it. I guess that it mostly works correctly even for partially cloned repositories.

@alfonsarmi  @skysb Can you share the git commands you use to clone your repository?

skysb:
@jairbubbles 
The git clone command that I am currently using is:

skysb:
@jairbubbles 

The gitclone command I am currently using is

`git clone --filter=blob:limit=1m  <repo_url> <folder>`

Using libgit2sharp, just doing a 

`Repository repo = new Repository(folder); `

raises the exception stated above, **but only in linux**, and not in windows.
I am not trying to run any actions on the repository, I am just trying to inspect the repository.

I hope this helps as I am pretty new to this!

Thanks!

jairbubbles:
@skysb Did you test it thoroughly? I guess it will fail as soon as we need to get file content on an object which is not yet retrieved. For instance when using `Lookup`.

alfonsarmi:
> @skysb Did you test it thoroughly? I guess it will fail as soon as we need to get file content on an object which is not yet retrieved. For instance when using `Lookup`.

Exactly the same problem in Linux, cannot "just read to inspect" the repository folder.

image

skysb:
> @skysb Did you test it thoroughly? I guess it will fail as soon as we need to get file content on an object which is not yet retrieved. For instance when using `Lookup`.

@jairbubbles 

Well, I can't test it until it lets me instantiate the repository right? 
I get the exact same error as @alfonsarmi .

jairbubbles:
Sorry @skysb I guess you only work on Linux on your side.

jairbubbles:
And my question was more:  "Did you test it thoroughly **on Windows**?"

alfonsarmi:
> And my question was more: "Did you test it thoroughly **on Windows**?"

Hi guys, 

Just in case it helps... @jairbubbles @skysb @ethomson 

I did test this thoroughly in Windows with a partially cloned repository.

So I could instanciate the Repository folder (new Repository(folder)), and also I could inspect the repository and navigate through the file system (tree) of a partially cloned repo (--filter=blob:limit=1m). 

The only difference is that those filtered files (more than 1 megabyte) have no blob content, but everything else works.

It is not the case on a Linux container, as I cannot even instanciate the Repository (new Repository(folder) raises the "unsupported extension name extensions.partialclone error").

I think it something with the NativeLibraries for Linux... but I'm not sure.

jairbubbles:
@alfonsarmi  @skysb  What version of git you're using on Linux to do the clone? It might explain the different behaviors.

When looking at https://git-scm.com/docs/repository-version#_partialclone it seems that "extensions.partialclone" should be set in the .git config of the repo.

skysb:
> > (I'm happy to add an unsupported extension error code.)
> 
> Cool but if it exists before the repo is initialized it won't fix what we try to achieve, ie inspecting partially cloned repository through libgit2. Some kind of read-only open would make sense?

This would actually help us a lot. Being able to at least have a read-only open to inspect it and it would be a significant improvement over just raising an error, as now there many people that are opting for partially cloned repos where the repo size is huge.

ethomson:
Sorry if my comment about writing corrupt repositories suggested that writing was the only concern.  Reading is a similar problem.  There's no read only view of a repo with sha256, for example.  So it's not like we can ignore extensions if you are opening a repo read only.  We would still need this mechanism for you to tell us what extensions you understand.

jairbubbles:
Yes @ethomson but in our case we're speaking only about the "partialclone" extension. The behavior regarding other extensions is another concern.

ethomson:
I'm sorry, this is just not a viable path forward.  Adding read-only support for repositories is not a trivial undertaking, and while it might solve your needs, it's not a particularly elegant solution nor a general purpose one.  Adding a mechanism where users tell us the extensions that they support has the advantage of being both easier and a necessary for the library anyway.

jairbubbles:
> Adding a mechanism where users tell us the extensions that they support has the advantage of being both easier and a necessary for the library anyway.

I really don't have any preference, I was just brainstorming around our use case. I don't exacly see what you have in mind and I don't see how as a client of the library I can "handle" an extension. In our case it would be just a way to tell "I know what I do, the repo is partially cloned that's fine with me, I'm just going to inspect it".

ethomson:
> In our case it would be just a way to tell "I know what I do, the repo is partially cloned that's fine with me, I'm just going to inspect it".

Ah!  Sorry, my bad.  Yes, that's precisely what I have in mind.  libgit2 needs a system where the user can say "hey, I know how to handle this extension" and then when we see a repo that has that extension, we continue.  Otherwise, we'll block as we do today.  At app startup, you'll just say "hey I know what to do with these extensions" and then we'll open any repositories that have that extension and you can do what you wish.

Eventually, we'll have some set of extensions that we support ourselves: new sha format, for example, won't need to be opt-in (though it could be opt _out_).  But you'll be able to opt in to extensions that we don't know anything about (like this one).

I'll see if I can tackle this over the weekend.

jairbubbles:
@ethomson Well that would all make us very happy here. I can look at the C# part if you'd like.

jairbubbles:
There is some goodness coming here: https://github.com/libgit2/libgit2sharp/pull/1908 😏

skysb:
@jairbubbles That sounds awesome and promising! Once approved, would there be a nuget package created for it too?
Thank you so much for the quick responses to this!

jairbubbles:
@skysb Yes as soon as it's merged in `libgit2` I'll make all the PR so that it's available here.

Please note that I also started a PR to make the code more friendly when using blobless repos: #1909

skysb:
@jairbubbles  youre a ⭐ !! Would #1908 mean we can download partially cloned repos (like blobless files) and initialize the repo or would we need to wait for #1909 for the feature?

Cheers

jairbubbles:
#1908 will be enough.

#1909 is a quality of life PR on top of it. Many methods throw when the blobs are not available, you can already try / catch but I feel like it's not very pretty as it generates many different exceptions depending on the called method (sometimes  `NullReferenceException`).

 It's just a draft, I'm also considering throwing a new `BlobNotAvailable` exception.

alfonsarmi:
This is fantastic guys! Many thanks!

@jairbubbles looks like the merge's checks have failed. Any plan to get this fixed any time soon?

Thanks!

jairbubbles:
@alfonsarmi Yes I need to consume the new native binaries but they are not yet on nuget.org. I'll make a PR as soon as version 1.3 is tagged, should be this week-end.

skysb:
Hello! Any updates on the PR? should this discussion move to the other PR instead??

jairbubbles:
Hello @skysb libgit2 was not tagged, I will update to a commit containing the fix. As long as we're sill in pre-release I think it's ok.

jairbubbles:
See https://github.com/libgit2/libgit2sharp.nativebinaries/pull/128

skysb:
@ethomson @jairbubbles Any updates on this?

Issue #1485, opened 2017-08-22
Describe does not support --match option


WalkerCodeRanger:

samartzidis:
It seems relatively easy to implement as the _GitDescribeOptions_ struct already has an unused _Pattern_ field. Please consider exposing this functionality.

PR #1906, opened 2021-08-24
Add support for git describe --match

thatfrankdev:
Addresses #1485


LibGit2Sharp/Core/Proxy.cs
```diff
@@ -691,6 +691,7 @@ public static void git_cred_free(IntPtr cred)
                     MaxCandidatesTags = 10,
                     OnlyFollowFirstParent = options.OnlyFollowFirstParent,
                     ShowCommitOidAsFallback = options.UseCommitIdAsFallback,
+                    Pattern = StrictUtf8Marshaler.FromManaged(options.Match)
```
thatfrankdev, line 694: I prefered using "Match" since it's the actual option's name in latest git versions, rather than libgit2's "Pattern".

PR #1908, opened 2021-09-11
Set / get supported extensions

jairbubbles:
- C# wrapper for https://github.com/libgit2/libgit2/pull/6031

(I'm not sure at all for the P/Invoke code)

### DIFF
--- a/LibGit2Sharp/Core/NativeMethods.cs
+++ b/LibGit2Sharp/Core/NativeMethods.cs
@@ -757,7 +757,7 @@ namespace LibGit2Sharp.Core
 
         // git_libgit2_opts(GIT_OPT_SET_EXTENSIONS, const char **extensions, size_t len)
         [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
-        internal static extern int git_libgit2_opts(int option, string[] extensions, int len);
+        internal static extern int git_libgit2_opts(int option, GitStrArray extensions);
 
         // git_libgit2_opts(GIT_OPT_GET_EXTENSIONS, git_strarray *out)
         [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
--- a/LibGit2Sharp/Core/NativeMethods.cs
+++ b/LibGit2Sharp/Core/NativeMethods.cs
@@ -754,6 +754,14 @@ namespace LibGit2Sharp.Core
         // git_libgit2_opts(GIT_OPT_GET_USER_AGENT, git_buf *buf)
         [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
         internal static extern int git_libgit2_opts(int option, GitBuf buf);
+
+        // git_libgit2_opts(GIT_OPT_SET_EXTENSIONS, const char **extensions, size_t len)
+        [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern int git_libgit2_opts(int option, string[] extensions, int len);
+
+        // git_libgit2_opts(GIT_OPT_GET_EXTENSIONS, git_strarray *out)
+        [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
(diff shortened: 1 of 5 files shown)


jairbubbles:
@skysb @alfonsarmi Once this is merged you'll be able to open blobless repositories by adding `GlobalSettings.SetExtensions("partialclone");` prior to opening any repositories.

Many methods can throw unexpectedly and my goal is to make it more confortable in 
 https://github.com/libgit2/libgit2sharp/pull/1909. Please let me know your thoughts and what would be the best for your use case.
 
 @bording Is there some place where doc could be added regarding the support of partially cloned repositories?

skysb:
Hi @jairbubbles 
That sounds great! Would a nuget be created so we can download and test it out, anytime soon?
In terms of #1909 I only intend to use it for blobless repositories or filtering out with patterns. Eitherways, in my use case, all I needed was to be able to instantiate the repository, as all data related to files and trees is already available in that. As long as I can read that high level data, we wouldn't need a more extensive research.

We had tried it on windows and it allowed us to get the tree data so it should be more than sufficient as of now. But Id still like to test it out to see if I'm missing any scenarios.

Cheers

jairbubbles:
Hi @skysb, I'm not maintainer so we just need to wait just a bit.

skysb:
@bording Hi, any ETC for this?

LibGit2Sharp/Core/NativeMethods.cs
```diff
@@ -812,6 +812,14 @@ private sealed class NativeShutdownObject : CriticalFinalizerObject
         // git_libgit2_opts(GIT_OPT_GET_USER_AGENT, git_buf *buf)
         [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
         internal static extern int git_libgit2_opts(int option, GitBuf buf);
+
+        // git_libgit2_opts(GIT_OPT_SET_EXTENSIONS, const char **extensions, size_t len)
+        [DllImport(libgit2, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern int git_libgit2_opts(int option, string[] extensions, int len);
```
bording, line 818: This P/Invoke looks problematic to me for a couple of reasons:

1. If you look at other examples of where `size_t` is being used, this is defined as `UIntPtr` in the signature. (see `git_odb_write` as example).

2. The larger issue is the string array. I know when individual strings are passed in, they always use the `StrictUtf8Marshaler`, so it seems likely that some additional marshaling is required here as well. This appears to the first instance of needing to call a function with a `char**` parameter, so there's no example to show to how to do this properly. Seems like we might need a new marshaler type to handle this.

The use of custom string marshalers was something that was in place before the start of my involvement with the project, so I don't have the historical context for why the default string marshaling isn't good enough. @ethomson Any change you can share some insight as why they are needed and what we need to do here?

bording (changes requested):

bording:
@jairbubbles Please rebase the PR to resolve the conflicts!

LibGit2Sharp/Core/NativeMethods.cs
jairbubbles: Thx @bording I tried something with `GitStrArrayManaged.BuildFrom` which uses `StrictUtf8Marshaler`internally. Let me know if you like it better, I'm doing my best here as I don't know much about marshalling 😅

jairbubbles:
@bording There seem to be some flakiness on some jobs. I don't know if it's related to my change. See https://github.com/libgit2/libgit2sharp/runs/4285262042?check_suite_focus=true

LibGit2Sharp/Core/NativeMethods.cs
bording: I overlooked `GitStrArrayManaged` when I was reviewing this for the first time. What you have now does appear to be more correct. I want to test a few things to convince myself that it's correct, and then I'll do another review pass.

bording (approved):

jairbubbles:
Thx @bording, let me know when you publish a package with the changes.

bording:

PR #1909, opened 2021-09-12
Simplify dealing with missing git objects

jairbubbles:
- Add `IsMissing` on `GitObject` to know if a git objects is missing without any exception involved
- Make some methods on `Blob` and `Tree` explicitly throw a `NotFoundException` (rather than a `NullReferenceException` later on when accessing the `null` handle)


jairbubbles:
Hi @ethomson  I'm playing a bit with blobless repositories. I was wondering if there is already a mechanism in libgit2 that would allow to download a git object (a blob in this case) on demand?

bording (approved):

bording:

Issue #1912, opened 2021-09-20
`Proxy.git_remote_connect()` hides exceptions


YuKitsune:
I've noticed when calling `Repository.ListRemoteReferences` with a URL that doesn't point to a valid git repo, I get a fairly unhelpful error:
```
LibGit2Sharp.LibGit2SharpException: this remote has never connected
   at LibGit2Sharp.Core.Ensure.HandleError(Int32 result)
   at LibGit2Sharp.Core.Proxy.git_remote_ls(Repository repository, RemoteHandle remote)
   at LibGit2Sharp.Repository.ListRemoteReferences(String url, CredentialsHandler credentialsProvider)
   at LibGit2Sharp.Repository.ListRemoteReferences(String url)
   at LibGit2SharpScratchpad.Program.Main(String[] args)
```

I did some digging and noticed in `Repository.ListRemoteReferences(string, CredentialsHandler)`, there is a call to `git_remote_connect`, then `git_remote_ls`. In `git_remote_connect`, if an exception is thrown (E.g: from the call to `Ensure.ZeroResult(res)`), it's caught and then ignored:
https://github.com/libgit2/libgit2sharp/blob/6329bea6d315f563d7107982bccbea4325947808/LibGit2Sharp/Core/Proxy.cs#L2163-L2176

It looks like this was introduced in this commit: https://github.com/libgit2/libgit2sharp/commit/6bc517f4d8103578e309ec7d8e206b6d9c6b2b3e

Because this doesn't re-throw the exception, `git_remote_ls` gets given a `null` transport, and thus throws the `this remote has never connected` error.

I would've expected this to re-throw the exception after disposing of the custom headers.

Seems like an easy fix and I'm happy to send through a PR, just wondering if this was intentional or a legitimate bug.

ethomson:
Yes, this looks like an error!

YuKitsune:
Thanks for checking @ethomson
I've raised a PR: https://github.com/libgit2/libgit2sharp/pull/1913

PR #1913, opened 2021-09-20
Fix git_remote_connect not throwing on non-zero result

YuKitsune:
Fixes https://github.com/libgit2/libgit2sharp/issues/1912

Couldn't really find a good spot to add tests here, open to suggestions.

### DIFF
--- a/LibGit2Sharp/Core/Proxy.cs
+++ b/LibGit2Sharp/Core/Proxy.cs
@@ -2172,6 +2172,7 @@ namespace LibGit2Sharp.Core
             catch (Exception)
             {
                 customHeaders.Dispose();
+                throw;
             }
         }
 


LibGit2Sharp/Core/Proxy.cs
```diff
@@ -2172,6 +2172,7 @@ public static unsafe void git_remote_connect(RemoteHandle remote, GitDirection d
             catch (Exception)
             {
                 customHeaders.Dispose();
+                throw;
```
A-Ovchinnikov-mx, line 2175: Hmm, how about just wrapping `customHeaders` in a `using` block?

```cs
using (var customHeaders = new GitStrArrayManaged())
{
    int res = NativeMethods.git_remote_connect(remote, direction, ref remoteCallbacks, ref proxyOptions, ref customHeaders.Array);
    Ensure.ZeroResult(res);
}
```

A-Ovchinnikov-mx (changes requested):

bording:
Looking at this code, I'm now wondering why it only tries to dispose `customHeaders` when an exception has occurred. It seems to me that it either should always be disposing, or it never actually needs to dispose.

Regardless, this PR is an improvement, so thanks for the contribution!

bording (approved):

bording:
LibGit2Sharp 0.27.0-preview-0182 includes this change.

PR #1919, opened 2021-10-17
Fix incorrect information in exceptions

OronDF343:
The parameters displayed in "Unknown fast forward strategy" and "Unknown merge analysis" were swapped. This change fixes that.


bording (approved):

bording:

PR #1921, opened 2021-11-04
Test native libraries using docker for multiple distros

arturcic:
Adds a test project that references the LibGit2Sharp.
Adds a `native-test` GitHub workflow that builds and the runs the test project to test the native libraries are loaded properly
This test project is build inside a docker image and the the docker image is run

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,17 +11,36 @@ env:
   DOTNET_NOLOGO: true
 jobs:
   build:
-    name: ${{ matrix.name }}
+    name: Build
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2.3.4
+        with:
+          fetch-depth: 0
+      - name: Install .NET SDK
+        uses: actions/setup-dotnet@v1.8.1
+        with:
+          dotnet-version: 6.0.x
+      - name: Build
+        run: dotnet build LibGit2Sharp.sln --configuration Release
+      - name: Upload packages
+        uses: actions/upload-artifact@v2.2.4
+        with:
+          name: NuGet packages
+          path: bin/Packages/
+          retention-days: 7
+  test:
+    name: Test / ${{ matrix.os }} / ${{ matrix.arch }} / ${{ matrix.tfm }}
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        include:
-          - os: windows-2019
-            name: Windows
-          - os: ubuntu-20.04
-            name: Linux
+        arch: [ amd64 ]
+        os: [windows-2019, macos-10.15]
+        tfm: [ net472, netcoreapp3.1, net5.0, net6.0 ]
+        exclude:
           - os: macos-10.15
-            name: macOS
+            tfm: net472
       fail-fast: false
     steps:
       - name: Checkout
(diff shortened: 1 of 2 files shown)


.github/workflows/native-test.yml
```diff
@@ -0,0 +1,54 @@
+name: native-test
+on:
+  workflow_dispatch:
```
arturcic, line 3: This workflow is run only manually - you can specify the version of the Native libraries you want to test

.github/workflows/native-test.yml
```diff
@@ -0,0 +1,54 @@
...
+        if: matrix.arch == 'arm64'
+      - name: Setup Docker Buildx
+        uses: docker/setup-buildx-action@v1
+        with:
+            install: true
+      - name: Build Docker Image
+        run: |
+            [[ ${{ matrix.sdk }} = '3.1' ]] && target="netcoreapp${{ matrix.sdk }}" || target="net${{ matrix.sdk }}"
+
+            docker buildx build -t native-test --output type=docker -f Dockerfile \
```
arturcic, line 46: We build the docker image that later is run to prove the native lib is loaded properly

Dockerfile
```diff
@@ -0,0 +1,11 @@
+ARG DISTRO
+ARG SDK
+FROM gittools/build-images:$DISTRO-sdk-$SDK
+COPY . ./
+
+ARG TARGET
+ARG VERSION
+
+RUN dotnet build ./LibGit2Sharp.LoadTest -p:TargetFrameworks=${TARGET} --output run -p:NativeBinaries=${VERSION}
```
arturcic, line 9: Build the test console for the appropriate target framework and version of the native libraries

LibGit2Sharp.LoadTest/Program.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        // get the path to the git repo
+        do
+        {
+            gitRepoPath = Path.Combine(path, ".git");
+            path = Directory.GetParent(path)!.FullName;
+        } while (!Directory.Exists(gitRepoPath));
+
+        using var repo = new Repository(gitRepoPath);
+        Console.WriteLine(GlobalSettings.Version.LibGit2CommitSha);
```
arturcic, line 21: if the native library is loaded properly these consoles will output the libgit2 sha and the branch name. If you want to have a different logic here, I can adjust

LibGit2Sharp/LibGit2Sharp.csproj
```diff
@@ -29,8 +29,12 @@
     <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" /> <!-- Needed for T4 generation -->
   </ItemGroup>
 
+  <PropertyGroup>
+    <NativeBinaries Condition="'$(NativeBinaries)'==''">[2.0.315-alpha.0.7]</NativeBinaries>
```
arturcic, line 33: This is needed to be able to set the version dynamically during `dotnet build`

arturcic:
@bording may you have a look on this PR?

bording:
This PR has some interesting ideas, but it's clear that it's been ported over from another repo (GitVersion?), so it has things that we don't need if we're actually part of this repo.

Thoughts while looking at the PR:
- There's no need for a separate solution. The test project should be part of the main solution.
- There's no need to customize the version of the native binaries package. It should just run against whatever version the main project is referencing.
- Do we even need a separate load check project? I'm wondering if it would be a good idea to instead run the entire test suite, triggering whenever a CI workflow is run.

Thoughts?

arturcic:
> This PR has some interesting ideas, but it's clear that it's been ported over from another repo (GitVersion?), so it has things that we don't need if we're actually part of this repo.

That is true, I use the same docker images to test `GitVersion` (including LibGit2Sharp part of GitVersion). but the Docker images are just the base images of each distro, with the dotnet sdk/runtime native dependencies for each distro, and on top of that the .net sdk/runtime installed depending on the selected version. So basically the images contain only dotnet and the dependencies needed to run dotnet. There is no GitVersion specific packages/code [there](https://github.com/GitTools/build-images).

arturcic:
@bording I reworked the PR. Now the tests are run inside of docker containers for several distros and .net versions. Unfortunately I had to disable te `arm64` jobs as they run for quite a long of time. If you don't mind we keep only the `amd64` jobs for now I can later check how to enable in another PR.

.github/workflows/ci.yml
```diff
@@ -58,3 +58,47 @@ jobs:
...
+          distro: centos.7
+      fail-fast: false
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2.3.4
+        with:
+          fetch-depth: 0
+
+      - name: Setup QEMU
+        uses: docker/setup-qemu-action@v1
```
bording, line 89: What's the benefit of using this action vs. the approach used in the native binaries workflow?

.github/workflows/ci.yml
```diff
@@ -58,3 +58,47 @@ jobs:
...
+        with:
+          image: tonistiigi/binfmt:latest
+          platforms: arm64
+        if: matrix.arch == 'arm64'
+      - name: Setup Docker Buildx
```
bording, line 94: What does this do for us vs. just using `docker buildx` directly? That seemed to work okay for the native binaries workflow.

For that matter, why is here at all if we aren't building an image in this workflow at all now?

.github/workflows/ci.yml
```diff
@@ -58,3 +58,47 @@ jobs:
...
+        uses: docker/setup-buildx-action@v1
+        with:
+            install: true
+
+      - name: Test with docker
+        run: |
+            [[ ${{ matrix.sdk }} = '3.1' ]] && target="netcoreapp${{ matrix.sdk }}" || target="net${{ matrix.sdk }}"
```
bording, line 101: What does this line do?

.github/workflows/ci.yml
```diff
@@ -58,3 +58,47 @@ jobs:
...
+
+            test_command="dotnet test LibGit2Sharp.sln --configuration Release -p:TargetFrameworks=$target -p:ExtraDefine=LEAKS_IDENTIFYING -p:GeneratePackageOnBuild=false"
```
bording, line 103: Why are you passing in `TargetFrameworks` like this instead of just using the `--framework` parameter?

.github/workflows/ci.yml
arturcic: Actually you are right, we can remove this additional step as the buildx is already installed. This was from the time I was playing with it

.github/workflows/ci.yml
arturcic: In the matrix we have only the digits representing the version number: 3.1, 5.0, 6.0, but the target to build has either `netcoreapp` or `net` prefix depending on the version. It's a way to express the `target = sdk == 3.1 ? netcoreapp3.1 : net5,0` ternary operation.

.github/workflows/ci.yml
arturcic: Unfortunately if we try to use the --framework it will not work in docker with .net50 only installed. It will try to find a .net6.0 sdk as it's in the list of target frameworks.

What I do is programatically set the target framework property to the exact one I need inside the specific docker image. So it's needed

.github/workflows/ci.yml
arturcic: I will use the approach in native binaries to make it consistent even though it does the same thing, it installs the qemu binaries to the Docker host. I prefer this one instead of the other one as this one allows to install only the arm64 version of qemu binaries, and it's an official docker GitHub action. I will keep it for now, but if you want I can change this one to the same in native binaries, (or I can change there to use the docker action instead) whichever you prefer

.github/workflows/ci.yml
arturcic: Have a look on this run https://github.com/libgit2/libgit2sharp/actions/runs/1456731650 this one uses `--framework` instead of `-p:TargetFrameworks`, and all `.net5.0` build fail. I will revert back to `-p:TargetFrameworks`

bording (approved):

bording:
Thanks for the PR. Eventually I think want to move to use docker images that are in this repo to give a bit more control. For example, I'd prefer to always use the latest SDK, and then have additional runtimes installed as needed for the tests.

That can be done later, and this is a great first step!

PR #1926, opened 2021-11-22
Add support to pass "initial head" on repo init

zentron:
By default when initializing a new repository, the initial head is set to be whatever the globally configured value from `init.defaultBranch`. LibGit2Sharp was already exposing the `InitialHead` pointer to pass a value down to `libgit2` however it was not being exposed through the managed classes.
This change introduces an `InitOptions` class to provide this optional parameter.

### DIFF
--- a/LibGit2Sharp/InitOptions.cs
+++ b/LibGit2Sharp/InitOptions.cs
@@ -1,0 +1,14 @@
+namespace LibGit2Sharp
+{
+    /// <summary>
+    /// Optional parameters when invoking Init.
+    /// </summary>
+    public sealed class InitOptions
+    {
+        /// <summary>
+        /// Use the specified name for the initial branch in the newly created repository.
+        /// If not specified, fall back to the default name
+        /// </summary>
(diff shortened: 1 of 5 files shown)


zentron:
Resolves: https://github.com/libgit2/libgit2sharp/issues/1929

LibGit2Sharp/InitOptions.cs
```diff
@@ -0,0 +1,14 @@
+namespace LibGit2Sharp
+{
+    /// <summary>
+    /// Optional parameters when invoking Init.
+    /// </summary>
+    public sealed class InitOptions
```
A-Ovchinnikov-mx, line 6: Hi @zentron , may I ask you to have a look at this commit in our fork and consider adopting changes from it?

https://github.com/mendix/libgit2sharp/commit/d6e7723820cc4cd5c9abe006710c80a20d41fb9c

More specifically, it might be great to include both `WorkdirPath` and `IsBare` into the `InitOptions` class.

Thanks!

PR #1936, opened 2022-01-03
Throw NotFoundException if trees are missing when computing diff

jairbubbles:


bording:
Is there a test you could add for this change?

jairbubbles:
> Is there a test you could add for this change?

Doing it right away.

bording (approved):

bording:

PR #1941, opened 2022-01-23
Added ObjectDatabase.MergeTrees() 

frindler:
This wraps the libgit2 function git_merge_trees() and so allows one to merge trees directly (specifying the ancestor tree to use in the 3-way merge). This functionality currently does not seem to be possible in libgit2sharp since the merge methods always take commits and _do not allow to specify the ancestor_. So, for instance, if I want to compute an octopus-merge ancestor and then iteratively merge N commits relative to this ancestor, then I cannot currently do this. With MergeTrees() one can. Later, one may write the resulting index to a tree via Index.WriteToTree() and create a commit with ObjectDatabase.CreateCommit().
Notes:
1) I added MergeTrees() to ObjectDatabase, not directly on Repository, because it seemed more low-level.
2) Internally I use ObjectHandle for trees (not a new "TreeHandle" via git_tree_lookup) - this seems to be how it's done everywhere else that uses Tree and it doesn't seem to matter in any case since libgit2 treats trees as objects.
3) I added 5 unit tests, which are basically the same unit tests as the existing merge ones for commits (I copied & adapted them), but with merging on the level of commits replaced with merging on the level of the corresponding trees.

### DIFF
--- a/LibGit2Sharp.Tests/MergeFixture.cs
+++ b/LibGit2Sharp.Tests/MergeFixture.cs
@@ -903,6 +903,100 @@ namespace LibGit2Sharp.Tests
         }
 
         [Fact]
+        public void CanTreeMergeTreeIntoSameTree()
+        {
+            string path = SandboxMergeTestRepo();
+            using (var repo = new Repository(path))
+            {
+                var master = repo.Branches["master"].Tip;
+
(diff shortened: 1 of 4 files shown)


A-Ovchinnikov-mx:
I would very much like to see it merged, as it turns out the default merge doesn't use an Octopus strategy, and thus in some circumstances it fails to determine a correct base. For now I'll cherry-pick this into our fork though.

PR #1957, opened 2022-04-01
Update LibGit2Sharp.NativeBinaries to 2.0.315

bording:
This update is libgit2 [1.4.2](https://github.com/libgit2/libgit2/releases/tag/v1.4.2).


bording:
Looks like libgit2 1.4.x has introduced some changes that we'll need to react to. Lots of test failures!

Balkoth:
I went in and manually built the NativeBinaries from libgit2 v1.5.1 sources. If i upgrade to my homebrew package only the test for `CanFetchCustomRefSpecsIntoAnEmptyRepository` fails. Someone with knowledge has to step in and see whats wrong here, as i can not really figure out what it's doing there.

PR #1959, opened 2022-04-06
Bugfix for .NET Framework using native libraries

Jamesthe1:
The section for .NET Core was left open for .NET Framework. As such, any build using .NET Framework would try to access NativeLibraries when it did not exist. The IF preprocessor statements have been swapped to fix this, and an additional build target for Framework 4.7.2 has been added.

### DIFF
--- a/LibGit2Sharp/Core/NativeMethods.cs
+++ b/LibGit2Sharp/Core/NativeMethods.cs
@@ -66,9 +66,7 @@ namespace LibGit2Sharp.Core
             return Path.Combine(nativeLibraryDir, libgit2 + Platform.GetNativeLibraryExtension());
         }
 
-#if NETSTANDARD
-        private static bool TryUseNativeLibrary() => false;
-#else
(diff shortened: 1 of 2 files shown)


bording:
I'm sorry, but the changes in this PR are not correct. This project targets `netstandard2.0` and `netcoreapp3.1`. `NativeLibrary.SetDllImportResolver` is only available in .NET Core 3.1 and newer, so the `netstandard2.0` assembly cannot attempt to use it.

#ifdefs affect compilation, not runtime, so when compiling the `netstandard2.0` assembly, we correctly exclude the `NativeLibrary.SetDllImportResolver` that would otherwise break compilation.

When you're running on .NET Framework, you are using the `netstandard2.0` assembly, so `NativeLibrary.SetDllImportResolver`, which doesn't exist on .NET Framework, is not called.

Thanks for the PR, but I'm going to close this since this is not a change that makes sense to merge.

PR #1980, opened 2022-08-08
Add NuGet package badge

mungojam:
I found it quite tricky to be sure of the nuget package name that I needed to add to my project. I generally look for the badge. I now see there is a link already, but I didn't spot it initially


bording:
Thanks!

PR #2006, opened 2023-01-14
Use expression instead of file for NuGet license 

mus65:
You can't set both and expressions are better because it allows tools to automatically determine the license.


bording:
Thanks for the PR, but this is not something I'm interested in changing at this time.

I find the usage of the expressions problematic because they do not actually include the full license text, for example the copyright statement.

mus65:
@bording Please reconsider this...

- NuGet specifically recommends ["If you're licensing the package under a common license, like MIT or BSD-2-Clause, use the associated SPDX license identifier."](https://learn.microsoft.com/en-us/nuget/reference/nuspec#license)
- The copyright is already included as separate metadata, so it's not an issue
- The full license text is the same anyway and NuGet automatically links to it on the website and in Visual Studio.
- some more background on why the expression is useful: we are automatically creating an SBOM from our package references in the CycloneDX format. The missing License Expression means we can't automatically determine the license and have to manually hardcode the license for LibGit2Sharp.

PR #2011, opened 2023-02-05
Add support for git push with custom headers

evolvedlight:
This is needed for pushing changes to bitbucket server repos


JackLiang5 (approved):

LibGit2Sharp/Network.cs
```diff
@@ -376,6 +376,7 @@ public virtual void Push(Remote remote, IEnumerable<string> pushRefSpecs, PushOp
                                           PackbuilderDegreeOfParallelism = pushOptions.PackbuilderDegreeOfParallelism,
                                           RemoteCallbacks = gitCallbacks,
                                           ProxyOptions = new GitProxyOptions { Version = 1 },
+                                          CustomHeaders = pushOptions.CustomHeaders?.Length > 0 ? GitStrArrayManaged.BuildFrom(pushOptions.CustomHeaders) : new GitStrArrayManaged()
```
JackLiang5, line 379: Is null better than new GitStrArrayManaged() here？

TackyTortoise:
Without this it is also seemingly impossible to push to Azure repository using a personal access token. I wonder why it was exposed in FetchOptions but not in PushOptions.

LibGit2Sharp/Network.cs
TackyTortoise: The CustomHeaders field is of type GitStrArrayManaged, which is a struct type so non-nullable value type. So not without some other impacting modifications I think.

James-LG, line 379: `GitStrArrayManaged` needs to be disposed. I have a similar PR that does that here #2052 by following the pattern set by `FetchOptionsWrapper`.

evolvedlight:
Sorry, somehow I missed the notifications on the comment. Will close in favor of https://github.com/libgit2/libgit2sharp/pull/2052

PR #2020, opened 2023-03-07
Include process-wide search path in default search

### LINKED ISSUE #2019: Allow loading native dependencies from PATH environment variable
LibGit2Sharp should respect the system specific library search paths. Not doing so prevents a wide-range of deployment scenarios from working properly.

It seems like LibGit2Sharp assumes specific directory structures for dependencies, which might or might not match the actual application deployment needs. There is an option to set native library paths explicitly, but this requires hard-coding strings in the application code and makes it impossible to separate coding of individual components from application packaging and deployment.

Please make sure to respect library search paths, such as the PATH environment variable in Windows (and similar strategies for Linux, etc).

### Reproduction steps

Deploy `git2-106a5f2.dll` to a separate folder. Set PATH to include this folder. Call LibGit2Sharp methods.

### Expected behavior

The native dependency is loaded since PATH includes its location as per [Dynamic-link library search order](https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order).

### Actual behavior

LibGit2Sharp crashes with "The specified file was not found" exception.

0.26.2

Windows 10 / 11

glopesdev:
This PR adds process-wide search paths to the set of included default import search paths. This is considered part of the safe values for `DllImportSearchPath` as per [CA5393](https://learn.microsoft.com/th-th/dotnet/fundamentals/code-analysis/quality-rules/ca5393#how-to-fix-violations).

Fixes #2019

### DIFF
--- a/LibGit2Sharp/Core/NativeMethods.cs
+++ b/LibGit2Sharp/Core/NativeMethods.cs
@@ -7,7 +7,7 @@ using System.Runtime.InteropServices;
 using LibGit2Sharp.Core.Handles;
 
 // Restrict the set of directories where the native library is loaded from to safe directories.
-[assembly: DefaultDllImportSearchPaths(DllImportSearchPath.AssemblyDirectory | DllImportSearchPath.ApplicationDirectory | DllImportSearchPath.SafeDirectories)]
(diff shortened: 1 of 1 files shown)


glopesdev:
Actually it turns out that adding `UserDirectories` does not resolve this problem. Instead, the application should call [`AddDllDirectory`](https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-adddlldirectory) to specify the process wide DLL search path.

PR #2042, opened 2023-06-10
add Owner Validation switch

kdlslyv:
adds a switch to disable the Owner Validation, which (afaik) checks the filesystem permissions of the repository accessed.

this check imho increases security in already wacky edge-cases - just educate the user to use git with some caution. 

overall this check is a major annoyance on environments where you have little influence on that like Azure App Services.


bording:
Thanks for the PR. I ended up bring this feature in via #2093.

PR #2043, opened 2023-06-17
Set GitProxyOptions type to default to auto

jones-gareth:
Fix for #1429.  

Currently `GitProxyOptions.Type` is set to `GitProxyType.none` and it is not possible to use  libgit2sharp through a proxy.  By setting `GitProxyOptions.Type` to `GitProxyType.Auto` git config settings for `http.proxy` and `https.proxy` work as expected in libgit2sharp.

PR https://github.com/libgit2/libgit2sharp/pull/1689 also attempts to address this, but is still open.  However, it is not legal to have a field initializer in a struct.

PR https://github.com/libgit2/libgit2sharp/pull/1708 did a nice job of fixing this issue, but it was closed by the author and not merged.

It would be *really* nice to have this feature in libgit2sharp.

### DIFF
--- a/LibGit2Sharp/Commands/Fetch.cs
+++ b/LibGit2Sharp/Commands/Fetch.cs
@@ -75,7 +75,7 @@ namespace LibGit2Sharp
                     fetchOptions.CustomHeaders = GitStrArrayManaged.BuildFrom(options.CustomHeaders);
                 }
 
-                fetchOptions.ProxyOptions = new GitProxyOptions { Version = 1 };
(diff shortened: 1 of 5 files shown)


jones-gareth:
I've created a [NUGET package](https://www.nuget.org/packages/libgit2sharp-proxy-enabled/) from this PR.

bording:
Thanks for the PR. I've incorporated setting the proxy type to auto by default in my comprehensive PR that adds proxy support, #2065.

PR #2044, opened 2023-06-19
Include merge commits in the file history

emanuel-v-r:
Currently when we use file history, it doesn't show any merge commits, which could be misleading, as it might contain changes for this specific file.
Also, the popular git providers show merge commits when showing the history of a single file.
Looks like the changes  in  this PR could fix it (currenlty it simply skips when there is more than one parent).
Eventually if we want to keep the old behavior, we may want to pass some kind of flag to the FileHistory class, maybe ``includeMergeCommits=false``


emanuel-v-r:
I am closing this PR as the approach that you want to follow is a more flexible one and I am not actually sure that those changes are as simple to implement, in the way that I am afraid of users breaking the seriaization/deserialization due to such flexibility.
Probably that would require more extensive testing as well.

PR #2053, opened 2023-08-02
Modernized argument checks in several files by using `nameof()`

Lehonti:
There are many other such checks that remain to be converted, but I figured they would perhaps be too many to review, so I'm doing splitting it in bite-sized pull requests. If this one is merged I want to do more of them in the future.


Lehonti:
@bording any plans to merge this? It's an easy and quick way to improve the code

PR #2052, opened 2023-07-25
feat: Add CustomHeaders to PushOptions

James-LG:
Wires up CustomHeaders in PushOptions in the same vein as FetchOptions.

Libgit2 already supports CustomerHeaders on PushOptions, so might as well hook it up. Plus it's required for some git servers that require strange PAT token formats like Azure DevOps #2041.

This PR is similar to #2011 except it properly disposes the managed string array and it has some tests.

PR #2063, opened 2023-11-02
Added CustomHeaders to PushOptions

jhested:


bording:
I have merged #2052 so this shouldn't be needed.

PR #2064, opened 2023-11-04
Try out new binaries package

bording:


bording:
@ethomson This PR is trying out updating to libgit2 1.7.1 (previously on 1.6.4), but I'm seeing some potentially unexpected new behavior.

One of the tests I've already changed in this PR since it appears libgit2 is returning extensions in a new order.

However, the other tests are failing to due changes in the number of tags that are being fetched when the tag fetch mode is either not configured or set to auto.

I'm guessing this is related to https://github.com/libgit2/libgit2/pull/6539, but I can't tell if this is an expected behavior change or some sort of bug. In at least one of the failures, it seems like doing the equivalent thing in git itself matches the old behavior that the tests were written against, and not what I'm seeing with 1.7.1.

bording:
Aha, looks like I figured it out. Looks like the shallow clone support in 1.7.0 changed `git_fetch_options` to add the `depth` value.

After adding that, looks like everything is green now.

@ethomson It would probably be a good idea to update the [libgit2 v1.7.0](https://github.com/libgit2/libgit2/releases/tag/v1.7.0) release notes to add that info to the breaking changes section.

bording:
Closing this PR in favor of #2066

PR #2065, opened 2023-11-26
Add proxy options

bording:
This PR adds a `ProxyOptions` class that can be used to specify proxy options to all of the relevant places where a proxy could be needed.

`ProxyType` is set to `Auto` by default, so LibGit2Sharp will now respect any proxy settings configured in git. `ProxyType` can be set to `None` to disable using the repo's proxy settings. `Specified` let's you specify a a proxy URL. There are also callbacks to deal with credentials and certificate verification.

As part of these changes, `CloneOptions` and `SubmoduleUpdateOptions` have been updated to have all fetch-related options be exposed via a `FetchOptions` property and not a confusing mix of some being directly on the class and others being only in an existing `FetchOptions` property.

I haven't added any proxy-specific automated tests here primarily because I don't currently know of a good way to have a proxy server spun up in GitHub Actions that I can reliably use on all platforms that need to be tested on.

For now, the manual testing I've done will have to be enough!

### DIFF
--- a/LibGit2Sharp.Tests/CloneFixture.cs
+++ b/LibGit2Sharp.Tests/CloneFixture.cs
@@ -567,10 +567,8 @@ namespace LibGit2Sharp.Tests
             const string url = "https://github.com/libgit2/TestGitRepository";
 
             const string knownHeader = "User-Agent: mygit-201";
-            var cloneOptions = new CloneOptions()
-            {
-                FetchOptions = new FetchOptions { CustomHeaders = new String[] { knownHeader } }
-            };
+            var cloneOptions = new CloneOptions();
+            cloneOptions.FetchOptions.CustomHeaders = new string[] { knownHeader };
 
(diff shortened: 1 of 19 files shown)


JonWillis-Tote:
Hi, This was my previous code, but this PR as mentioned is breaking. 
Is there documentation on how to use the new proxy feature so it works

```
 var cloneOptions = new CloneOptions();
 var usernamePasswordCredentials = new UsernamePasswordCredentials { Username = _githubOptions.Username, Password = _githubOptions.Token };
 cloneOptions.CredentialsProvider = (_url, _user, _cred) => usernamePasswordCredentials;
 Repository.Clone("https://github.com/Jon/MyRepo.git", directory, cloneOptions);
```

As CredentialsProvider has been removed, I did try replacing it with this, but it failed.

```
cloneOptions.FetchOptions.ProxyOptions.CredentialsProvider = (_url, _user, _cred) => usernamePasswordCredentials;
```

PR #2093, opened 2024-05-07
Make owner validation configurable

### LINKED ISSUE #2036: Expose the repository directory ownership validation option
You can now disable the repository directory ownership validation by using the `GIT_OPT_SET_OWNER_VALIDATION` with `git_libgit2_opts()`: https://github.com/libgit2/libgit2/pull/6266

This option is not yet exposed in the C# bindings, e.g. in the `GlobalSettings` static class.

Thanks.

ltrzesniewski:
This adds a `GlobalSettings.OwnerValidation` property which wraps `git_libgit2_opts(GIT_OPT_[GET/SET]_OWNER_VALIDATION)`.

I needed to disable this (I'm using a safe environment), and I had to do this through reflection, but I suppose other people may have the same problem so exposing the feature in the library would be better. 🙂 

Closes #2036

**Edit:** I'm an idiot, I forgot to check whether someone else submitted a PR for this, so this duplicates #2042, sorry 🤦  
I'm leaving this open so you can choose the implementation, and this one also lets you retrieve the current option value.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ Thumbs.db
 *.ncb
 *.suo
 .vs/
+.idea/
 *.sln.ide/
 *.tlb
 *.tlh
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,6 @@ Thumbs.db
 *.ncb
 *.suo
 .vs/
-.idea/
 *.sln.ide/
 *.tlb
 *.tlh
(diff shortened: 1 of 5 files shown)


bording (approved):

bording:
@ltrzesniewski Thanks for the PR. Of the open PRs for this feature, yours was was the most complete (it even included a test!) so I decided to it to get this feature in.

I made a few tweaks, primarily moving away from a property and instead having separate get/set methods. Overall I think separate methods make more sense when there are separate libgit2 methods that need to be called.

PR #2099, opened 2024-06-04
Fix empty folder issue in Repositoy.Worktrees.Add

enriqueraso:
This PR is fixing an issue when creating a worktree. It creates an empty folder with .git file only. 0.26.2 was the Latest version that creating a worktree works well and it was using the GIT_CHECKOUT_FORCE as checkout strategy.

* Set GIT_CHECKOUT_FORCE as checkout strategy when creating a worktree.
* Update current and add unit tests in WorktreeFixture.

### DIFF
--- a/LibGit2Sharp.Tests/WorktreeFixture.cs
+++ b/LibGit2Sharp.Tests/WorktreeFixture.cs
@@ -3,8 +3,6 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


bording:
@enriqueraso Thanks for the PR. I rebased it and made a few tweaks. I also decided to use `GIT_CHECKOUT_SAFE` instead of `GIT_CHECKOUT_FORCE` since that aligns better with the libgit2 default setting.

bording (approved):

enriqueraso:
Thxs @bording

PR #2113, opened 2024-07-19
Target netstandard2.0 rather than net472

### LINKED ISSUE #2035: LibGit2Sharp should target netstandard
[PR 2026](https://github.com/libgit2/libgit2sharp/pull/2026/files) changed the lib from targeting netstandard2.0 to net6.0 and netframework472. What was the rationale behind this decision? This means I'm unable to include LibGit2Sharp in my netstandard2.0 library's package references without it throwing a warning.

Targeting netstandard2.0 will encompass netframework472, net6, and net7.

### Reproduction steps
Create a netstandard2.0 library project with `<PackageReference Include="LibGit2Sharp" Version="0.27.2" />`

### Expected behavior
Library compiles without warnings.

### Actual behavior
Upon building my netstandard2.0 library which references LibGit2Sharp, I receive the following warning:

`warning NU1701: Package 'LibGit2Sharp 0.27.2' was restored using '.NETFramework,Version=v4.6.1, .NETFramework,Version=v4.6.2, .NETFramework,Version=v4.7, .NETFramework,Version=v4.7.1, .NETFramework,Version=v4.7.2, .NETFramework,Version=v4.8, .NETFramework,Version=v4.8.1' instead of the project target framework '.NETStandard,Version=v2.0'. This package may not be fully compatible with your project.`

0.27.2

Ubuntu-20.04, .NET 6 app which links to netstandard2.0 library; netstandard2.0 library links to LibGit2Sharp.

louis-z:
Targeting `netstandard2.0` will make the library more versatile.

According to [Microsoft's Cross-platform targeting recommendations](https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting):

> Most general-purpose libraries don't need APIs outside of .NET Standard 2.0. .NET Standard 2.0 is supported by all modern platforms and is the recommended way to support multiple platforms with one target.

Closes #2035

### DIFF
--- a/LibGit2Sharp/Core/NativeMethods.cs
+++ b/LibGit2Sharp/Core/NativeMethods.cs
@@ -66,7 +66,7 @@ namespace LibGit2Sharp.Core
             return Path.Combine(nativeLibraryDir, libgit2 + Platform.GetNativeLibraryExtension());
         }
 
-#if NETFRAMEWORK
+#if NETSTANDARD
         private static bool TryUseNativeLibrary() => false;
 #else
         private static bool TryUseNativeLibrary()
--- a/LibGit2Sharp/GlobalSettings.cs
+++ b/LibGit2Sharp/GlobalSettings.cs
@@ -29,7 +29,7 @@ namespace LibGit2Sharp
 
             nativeLibraryPathAllowed = netFX || netCore;
 
-#if NETFRAMEWORK
+#if NETSTANDARD
             if (netFX)
             {
                 // For .NET Framework apps the dependencies are deployed to lib/win32/{architecture} directory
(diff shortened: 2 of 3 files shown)


bording:
Thanks for the PR, but I'm going to be taking this change.

LibGit2Sharp is not a typical library and must know what underlying platform it's being run on to ensure that the native libraries are loading properly, so it very intentionally doesn't target `netstandard2.0` at this point.

The changes you've made in this PR would result in a non-functional assembly if you tried to use the `netstandard2.0` target on something like an earlier version of .NET Core, for example.

louis-z:
> Thanks for the PR, but I'm going to be taking this change.
> 
> LibGit2Sharp is not a typical library and must know what underlying platform it's being run on to ensure that the native libraries are loading properly, so it very intentionally doesn't target `netstandard2.0` at this point.
> 
> The changes you've made in this PR would result in a non-functional assembly if you tried to use the `netstandard2.0` target on something like an earlier version of .NET Core, for example.

I kind of suspected as much, but since all tests were green I thought maybe it was fine after all...

@bording, do you have any idea when the issue will be resolved? This is kind of a blocker for us right now.

bording:
> @bording, do you have any idea when the issue will be resolved? This is kind of a blocker for us right now.

What issue? From my perspective there is no issue to be solved. How is this blocking you?

PR #2120, opened 2024-08-08
Add support for long file paths in CheckoutOptions

DenisKudelin:
## Summary
Adds support for long file paths in `CheckoutOptions` by introducing the `LongPaths` property and updating `CheckoutStrategy` with the `GIT_CHECKOUT_LONGPATHS` flag.

## Changes
- New property `LongPaths` in `CheckoutOptions`.
- Added `GIT_CHECKOUT_LONGPATHS` to `CheckoutStrategy`.

## Motivation
Enhances compatibility with projects that have deeply nested directories or long file names, especially on Windows.

### DIFF
--- a/LibGit2Sharp.Tests/CloneFixture.cs
+++ b/LibGit2Sharp.Tests/CloneFixture.cs
@@ -629,5 +629,30 @@ namespace LibGit2Sharp.Tests
             var clonedRepoPath = Repository.Clone(url, scd.DirectoryPath, cloneOptions);
             Assert.True(Directory.Exists(clonedRepoPath));
         }
+
+        [Fact]
+        public void CanCloneWithLongPaths()
+        {
+            var scd = BuildSelfCleaningDirectory();
+
+            var cloneOptions = new CloneOptions
+            {
+                LongPaths = true
+            };
+
+            string clonedRepoPath = Repository.Clone("https://github.com/luiswolff/test-long-file-path.git", scd.DirectoryPath, cloneOptions);
+
+            using (var repo = new Repository(clonedRepoPath))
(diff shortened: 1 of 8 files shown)


DrewScoggins:
@bording could you take a look at this PR and help get it merged?

Ogglas:
@bording We really need this bug fixed. Please prioritize it if possible

LoopedBard3:
@Ogglas, I was able to workaround this by setting the longpath option in the global git config. Presuming a global git config already being present on the machine at %USERPROFILE%\.gitconfig, we create a temp repo (`new Repository(Repository.Init("temprepo"))`) and then set the longpath setting in that global config with `temp_repo.Config.Set("core.longpaths", true, ConfigurationLevel.Global);`. With that set, the clone should use the global setting as long as it is not set at a more local level.

bording:
As far as I can tell, this PR isn't actually doing anything. The support for long paths on Windows that was added to libgit2 in https://github.com/libgit2/libgit2/pull/5857 only supports the `core.longpaths` configuration setting.

Furthermore, the `GIT_CHECKOUT_LONGPATHS` setting that's been added to the `CheckoutStrategy` enum [doesn't seem to exist at all in libgit2](https://github.com/libgit2/libgit2/blob/c6111ec06cc0b0126cd67bdab78f8091e4cfaa10/include/git2/checkout.h#L107-L200).

`(1 << 24)` is an entirely different option, `GIT_CHECKOUT_DRY_RUN`.

The good news is that LibGit2Sharp has been respecting the `core.longpaths` setting since at least 0.27.0, so you can already use long paths without needing any other changes.

Issue #2159, opened 2025-06-25
LibGit2SharpException loses information in HandleError


asinghmodesttree:
In `Ensure.HandleError` function, since we do not have an exception for handling network errors, when fetch fails and the exception of type LibGit2SharpException exception is thrown, its thrown with arguments (Message, GitErrorCategory) but the exception doesn't have constructor that take an argument of type GitErrorCategory and the closest match is `object params` which is to handle string format arguments for messages, so the GitErrorCategory information is lost.

Ensure.cs:151 `new LibGit2SharpException(m, c)`

I am trying to understand if there is any technical reason it is done this way?

asinghmodesttree:
I didn't realize that GitErrorCategory is internal, closing this.

zentron:
Hi @asinghmodesttree .
This issue has bitten us in the past when the returned error string has curly braces around text. It seems fairly unlikely that the error message would have the `{0}` value required to be replaced with the `GitErrorCategory` code.

As a result we have had to put a fix for this in our fork.

Much appreciated if you could merge [this PR](https://github.com/libgit2/libgit2sharp/pull/2163) to avoid this divergence.

PR #2163, opened 2025-07-19
Do not use `LibGit2SharpException` overload for unknown message

zentron:
# Background
The `LibGit2SharpException` class has a constructor overload that takes two parameters and attempts to call `String.Format` to convert the values into a single string.

Unfortunately if the provided message contains text between curly braces then the string.format throws an exception.

Related PR https://github.com/OctopusDeploy/libgit2sharp/pull/10

## Result
Since this message could have any value, it seems risky to assume that the provided string will have a `{0}` that will accept the `GitErrorCategory` as a replacement value. In fact it seems rather unlikely that this would typically apply at all. As a result this handler would ignore the error category in the best case, or throw an exception in the worst case.

## Alternative
I'd also be happy to ignore the category entirely since its likely that it is of type `GitErrorCategory.Unknown` anyway

Alternative snippet
```
exceptionBuilder = (m, c) => new LibGit2SharpException(m);
```


zentron:
@bording  this is related to a closed issue https://github.com/libgit2/libgit2sharp/issues/2159

PR #2173, opened 2025-10-30
Bring the GitRemoteCallbacks structure in sync with libgit2

### LINKED ISSUE #2161: "invalid version 0 on git_proxy_options" thrown when natives are built from source
Hi, I'm currently trying to make a custom build that uses `libssh2` instead of `exec` for SSH, since I don't want OpenSSH windows to spawn as my program makes requests. I changed this flag, then followed through the rest of the compilation steps both for native binaries and lg2s.

I then tried pulling, not on a proxy, using this code:

```cs
var fOptions = new FetchOptions ();
fOptions.Prune = UpdaterMod.settings.pruneOld;

var mOptions = new MergeOptions ();
mOptions.CommitOnSuccess = true;
mOptions.FileConflictStrategy = UpdaterMod.settings.onFileConflict;

var sig = new Signature (new Identity (modName, email), DateTimeOffset.Now);

try {
    string logHolder = "";
    Commands.Fetch (repo, remote.Name, refSpecs, fOptions, logHolder);

    if (logHolder.Length > 0)
        LogMsg (logHolder, LogMode.Warn);
}
catch (Exception e) {
    LogMsg ($"Could not fetch data from remote repo: {e.Message}", LogMode.Error);
    return;
}
```

## Expected behavior

The program should attempt to pull without trying to use a proxy.

## Actual behavior

The caught exception contains the message in the title.

## System information

Windows 10 10.0.19045

AMD Ryzen 5 3600

## Other useful information

This issue persists when the flag is switched back to exec

arrowd:
The marshaling was messed up due to missing extra field.

Fixes #2161

### DIFF
--- a/LibGit2Sharp/Core/GitRemoteCallbacks.cs
+++ b/LibGit2Sharp/Core/GitRemoteCallbacks.cs
@@ -33,10 +33,12 @@ namespace LibGit2Sharp.Core
 
         internal IntPtr transport;
 
-        private IntPtr padding; // TODO: add git_remote_ready_cb
+        internal IntPtr remote_ready; // TODO: add git_remote_ready_cb
 
         internal IntPtr payload;
 
         internal NativeMethods.url_resolve_callback resolve_url;
+
+        private IntPtr update_refs;
     }
 }


bording:
@arrowd I don't think you were looking at the right version of the libgit2 code when doing your investigations.

LibGit2Sharp is currently using v1.8.4, and that version of the `git_remote_callbacks` struct doesn't have the field you've added: https://github.com/libgit2/libgit2/blob/v1.8.4/include/git2/remote.h#L557-L658

arrowd:
The mentioned issue was talking exactly about using non-bundled libgit2 version.

I realize that this PR can't be merged as of now, but it will come in handy if libgit2sharp would bump its bundled libgit2.

bording:
> The mentioned issue was talking exactly about using non-bundled libgit2 version.

A non-bundled version still needs to be v1.8.4 to be compatible with LibGit2Sharp.

I'm going to go ahead and close this for now then.

PR #2182, opened 2026-03-16
Support libssh2 based callbacks

zentron:
This PR is not intended to be merged as-is

### DIFF
--- a/LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
+++ b/LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
@@ -1,13 +1,13 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks Condition="'$(TargetFrameworks)'==''">net472;net8.0;net9.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(TargetFrameworks)'==''">net10.0</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\LibGit2Sharp\LibGit2Sharp.csproj" />
-    <ProjectReference Include="..\NativeLibraryLoadTestApp\x86\NativeLibraryLoadTestApp.x86.csproj" Condition="'$(TargetFramework)' == 'net472'" ReferenceOutputAssembly="false" OutputItemType="TestAppExe" />
-    <ProjectReference Include="..\NativeLibraryLoadTestApp\x64\NativeLibraryLoadTestApp.x64.csproj" Condition="'$(TargetFramework)' == 'net472'" ReferenceOutputAssembly="false" OutputItemType="TestAppExe" />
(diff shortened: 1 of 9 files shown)


LibGit2Sharp.Tests/LibGit2Sharp.Tests.csproj
```diff
@@ -1,13 +1,13 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks Condition="'$(TargetFrameworks)'==''">net472;net8.0;net9.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(TargetFrameworks)'==''">net10.0</TargetFrameworks>
```
zentron, line 4: Bunch of changes to get it building locally

LibGit2Sharp/Core/SshExtensions.cs
```diff
@@ -0,0 +1,141 @@
+
```
zentron, line 1: This methods are required to be able to call down into the c libraries methods that are only available when using libssh2. The files in this class can be properly tidied up and split up before merge

LibGit2Sharp/GlobalSettings.cs
```diff
@@ -20,7 +20,7 @@ public static class GlobalSettings
 
         private static string nativeLibraryPath;
         private static bool nativeLibraryPathLocked;
-        private static readonly string nativeLibraryDefaultPath = null;
+        private static readonly string nativeLibraryDefaultPath = "/Users/robert/Development/Sandbox/libgit2/build";
```
zentron, line 23: Obviously not something to merge, just that helped testing against my local build of libgit2

LibGit2Sharp/ListRemoteOptions.cs
```diff
@@ -0,0 +1,26 @@
+using LibGit2Sharp.Handlers;
+
+namespace LibGit2Sharp;
+
+/// <summary>
+/// Options controlling ListRemote behavior.
+/// </summary>
+public sealed class ListRemoteOptions
```
zentron, line 8: New `ListRemoteOptions` required to get appropriate handlers on the network list operation

zentron: This can probably be put in as its own PR back into the upstream repo as its not really dependant on anything ssh

LibGit2Sharp/Network.cs
```diff
@@ -52,9 +52,34 @@ public virtual IEnumerable<Reference> ListReferences(Remote remote)
         {
             Ensure.ArgumentNotNull(remote, "remote");
 
-            return ListReferencesInternal(remote.Url, null, new ProxyOptions());
+            var options = new ListRemoteOptions()
```
zentron, line 55: There are a bunch of overloads in this network class to utilize the new ListRemoteOptions, these dont necessarily need to be implemented exactly like this

LibGit2Sharp/RemoteCallbacks.cs
```diff
@@ -285,6 +296,17 @@ private int GitCredentialHandler(
                 types |= SupportedCredentialTypes.Default;
             }
 
+            if (credTypes.HasFlag(GitCredentialType.SshKey))
```
zentron, line 299: To cater for the different credential callback for ssh commands these new enum values need to be catered for here

